PR #13, opened 2022-04-29
Directory Org Card Spacing

### LINKED ISSUE #3: Correct spacing behavior between listed organizations in the Directory
Currently, organizations listed in the directory have odd spacing under some conditions. When there is ample horizontal space, the organizations appear squished together, even if they have plenty of space to be more spread out. Organizations may also unexpectedly move to a new line when there is enough horizontal space to stay on the previous line.
Screen Shot 2022-04-11 at 1 04 11 PM

Jake3231:
This PR implements a fix for the incorrect spacing of org cards in the Directory. It turns out that the grid of cards was not taking up the full screen width, which is resolved here. Additionally, the spacing in each size class is refined so that items are properly spaced apart as the screen width changes.


Buchimi:
The spacing is good but the click boxes are still too big

Jake3231:
No longer with needed with the changes in #36 .

Issue #33, opened 2022-10-06
Missing placeholder Explore tab in navigation


Jake3231:
# Overview

As a user I should be able to select an Explore tab in the navigation to see recommended organizations.

# Proposal
In preparation for later changes, it'll be important to add a placeholder "Explore" button in the navigation bar alongside the Jupiter logo. For this change, the button does not need to be functional, but should appear in the correct place.

## Supplementary Information
This is the design that this originates from:
<img width="500" alt="Screenshot 2022-10-06 at 5 52 12 PM" src="https://user-images.githubusercontent.com/13911970/194432821-3e41791e-709f-484d-aa9a-ea749d228a1c.png">

Jake3231:
Verified fix with https://github.com/UTDNebula/jupiter/commit/63581a970bb41943d501165f09a18f29cb4a5993.

PR #35, opened 2022-10-07
Added "Explore" placeholder link to navbar

Master7130:
Fixes issue #33


Jake3231 (approved):
Looks good!

We may want to later consider from a design perspective what the correct alignment should be for the Explore button and the logo text.

Master7130:
> Looks good!
> 
> We may want to later consider from a design perspective what the correct alignment should be for the Explore button and the logo text.

Are you referring to how the logo text and the Explore button are slightly misaligned?
image

Jake3231:
Yes. This change does follow the expected design according to current spec, though we may want to consider changes to make this appear more aligned.

cc @Hanneisuru

Master7130:
> Yes. This change does follow the expected design according to current spec, though we may want to consider changes to make this appear more aligned.

Ok, I'll take a look at how I can fix that.

PR #36, opened 2022-10-08
Update directory to have filter sidebar and use nextJs image

RubenOlano:
Also created new components to be used on the directory page


Jake3231:
This looks really good! To be sure I understand, this PR also changes the implementation of the grid used in the directory, in addition to moving it into a reusable component?

Jake3231:
Additionally, if `react-simple-flex-grid` can be removed with this change then that may resolve some of the additional issues we've been running into.

RubenOlano:
Yeah, we can use tailwind for most of our styling needs. We can eventually migrate to not using any outside CSS and then remove the dependency if the team feels that that would be good

Jake3231 (changes requested):
Looks really good! One minor thing it that the new UI seems to have some unexpected behavior at lower display widths. With a few small changes this should be good to go.

Screenshot 2022-10-08 at 5 11 05 PM

RubenOlano:
Let me know if that looks a bit better. I added a border around each card so they would be easier to tell apart.

Jake3231 (changes requested):
Looks like responsiveness is improved with those changes. The design update does look nice, though for the scope of this change it would be best to refrain from major design changes, as it might be better to have a separate PR updating that to allow for more review and discussion.

components/OrgDirectoryTags.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+
+const mockTags = ['Tag'];
+const X = () => (
```
Jake3231, line 4: I'd consider adjusting this constant name to be something that could be more clear at a glance, such as `xButton`.

RubenOlano: I renamed it to XButton, because it won't allow me to use the element if it starts with a lowercase character for some reason. Should still be good regardless

Jake3231 (approved):

PR #41, opened 2022-10-12
Update tags to start at a baseline

RubenOlano:
Tags should move below a baseline rather than pushing elements up


RubenOlano:
Idk if I'm supposed to be able to assign the engineers to pull requests but I'm currently unable to

Jake3231 (approved):
Looks good!

Maybe we should consider the layout behavior from a design perspective - should tags be left aligned on a second row? cc @Hanneisuru

PR #42, opened 2022-10-17
Update navbar to only have either hamburger menu or nav buttons

RubenOlano:
Smaller screens hide the buttons, opting to show the hamburger menu
image
image


RubenOlano:
Didn't want to touch how the hamburger menu looks when clicked, but resizing works as expected

Jake3231 (approved):
Looks good!

This fixes the biggest issue with the behavior here, next we'll need to ensure that the additional nav options appear in some sort of secondary menu when clicking the hamburger button.

PR #55, opened 2022-10-23
Create placeholder for signin and logout buttons

RubenOlano:
Simple temporary set up until official auth flow is set up


Jake3231 (approved):
This looks really good! I think that as we get more UI built out we can move the "Sign Out" button into some sort of menu, though this should be really good to continue moving forward.

PR #57, opened 2022-10-23
Events placeholder

### LINKED ISSUE #9: Create Events placeholder on homepage
Currently, the Homepage includes just the banner images with the title "Jupiter" and a tagline. There needs to be a list of recommended events below that, as included in the design spec:
Screen Shot 2022-04-20 at 10 42 23 PM

Master7130:
Created EventCard component and added it to the homepage as a placeholder for the events section

Resolves #9

### DIFF
--- a/components/EventCard.tsx
+++ b/components/EventCard.tsx
@@ -5,10 +5,12 @@ type Props = { name: string };
 
 export const EventCard = ({ name }: Props) => {
   return (
-    <div className="pt-64 flex flex-col justify-center w-full h-full text-center">
-      <img
+    <div className="flex flex-col justify-center w-full h-full text-center">
+      <Image
         src="https://about.utdnebula.com/assets/images/nebula-logo-wordmark-dark-0b2813eedf7447691178ca82b5f53dce.png"
-        className="bg-gray-300 h-44"
+        height={200}
+        width={200}
+        className="bg-gray-300"
       />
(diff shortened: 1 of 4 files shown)


Jake3231 (changes requested):
That is looking pretty good! Looks like the last thing I needs is the "Events" title as shown in the Figma design:

Screenshot 2022-10-23 at 7 11 02 PM

RubenOlano:
I would also switch to using the NextJS image component for consistency as it works well with the vercel deploy

Master7130:
@RubenOlano I switched to NextJS Image but I'm having trouble resizing it. My code is in the image below. Right now the background fills up the whole screen. The image isn't being contained inside the div and just covers the whole screen. Any idea on how to fix this?

image

image

RubenOlano:
image
Something along these lines, contain and fill will just fill up its parent container which I think is being set to the full screen.

RubenOlano:
> @RubenOlano I switched to NextJS Image but I'm having trouble resizing it. My code is in the image below. Right now the background fills up the whole screen. The image isn't being contained inside the div and just covers the whole screen. Any idea on how to fix this?
> 
> image
> 
> image

image
You're gonna have to set the height/width manually. The `layout='fill'` and 'objectFit='contain'` makes it fill up it's parent component which is why its taking up the entire screen

PR #60, opened 2022-10-24
Update navbar logo

RubenOlano:
Should remain around the same size when making screens smaller


Jake3231 (approved):
Thanks for fixing that!

PR #64, opened 2022-10-24
Remove space at right side of navbar

Jake3231:
This PR changes the layout code for the navbar to resolve an unexpected space on the right side.


RubenOlano (approved):
Good catch on that issue

PR #65, opened 2022-10-24
44 additional sample organizations needed

felixjimenez4:
Added 5 additional clubs with some new tags.


RubenOlano (approved):
Looks good to me

Issue #59, opened 2022-10-24
Events Directory UI needed


Jake3231:
# Overview

As a user, I should be able to browse upcoming events across multiple organizations on the "Events" page. This will require creating a new page similar to the Directory page, but rather shows events. It includes a similar left sidebar as well as a vertically scrolling list of events.

# Proposal
This change should include non-functional UI which can be connected to a backend at a later time. Suggested steps are:

- [ ] Build vertically scrolling list
- [ ] Create enough sample events to show scrolling
- [ ] #61
- [ ] Add left-sidebar filtering UI

If this feature is largge and requires mutliple parts,

## Supplementary Information
![Screenshot 2022-10-24 at 10 53 01 AM](https://user-images.githubusercontent.com/13911970/197570383-d598d1b1-4c18-45fc-a673-e0f9838bbe5b.png)

This is the expected design, per the Figma document.

PR #68, opened 2022-10-25
Create events directory and create mock api and mock data to use

RubenOlano:
Addresses #59. Used mostly the same pattern as Directory to maintain consistency.


components/OrgDirectorySidebar.tsx
```diff
@@ -1,20 +1,6 @@
+import { DropDownIcon } from './../Dropdown';
```
Jake3231, line 1: Looks like this is causing a runtime error in the Directory:

```
./components/OrgDirectorySidebar.tsx:1:0
Module not found: Can't resolve './../Dropdown'
> 1 | import { DropDownIcon } from './../Dropdown';
  2 | import React from 'react';
  3 | 
  4 | const OrgDirectorySidebar = () => {

Import trace for requested module:
./pages/directory.tsx

https://nextjs.org/docs/messages/module-not-found
```

components/Event.tsx
```diff
@@ -0,0 +1,61 @@
...
+            alt={event.eventTitle}
+          />
+        </div>
+      </div>
+      <div className="col-start-2 col-end-[-2]">
+        <h4 className="text-xl font-bold text-center md:text-left">
+          {event.eventTitle}
+        </h4>
+        <p className="text-lg font-light text-center md:text-left">
+          {event.Description}
```
Jake3231, line 37: This may be nitpicky, though I wonder if the sample data properties should be lowercase to allow this to follow the convention of lowercase property names

Master7130:
Wouldn't it be better to have the "Events" text be aligned to the left of either the filters or the events section? 

image

Jake3231 (approved):
Overall looks great!

Jake3231:
> Wouldn't it be better to have the "Events" text be aligned to the left of either the filters or the events section?
> 
> image

That's a good point - that may be a broader design question that we should answer as it'll need to be consistent between the Directory and Events page.

RubenOlano:
I can work on aligning these two in a different PR if an issue is opened. I just kinda eyeballed the figma file but I agree that they should be aligned

Jake3231:
@Master7130 Does this change look good to you as well?

components/Event.tsx
```diff
@@ -0,0 +1,61 @@
+import React, { FC, useState } from 'react';
+import demoEvents from '../demoEvents.json';
+import Image from 'next/image';
+interface Props {
+  event: typeof demoEvents[0];
+}
+
+const Event: FC<Props> = ({ event }) => {
```
Master7130, line 8: Since the _event_ object is being used quite frequently shouldn't we be destructuring its different properties here?

Master7130:
> @Master7130 Does this change look good to you as well?

@Jake3231 Yeah other than the couple small details above it looks good to me. Also looking at the code reminded me about using [prettier to sort tailwindcss styles](https://tailwindcss.com/docs/editor-setup#automatic-class-sorting-with-prettier). This would be a good addition to keep our code more readable.

components/Event.tsx
RubenOlano: Yeah, I can implement a quick change for that

components/Event.tsx
RubenOlano: I think once we have a fully developed API, what we can do is instead of passing the whole object to the component, we can just loop through event ids and call the corresponding API to fetch that object from within the component. I just kinda made it like that for now so we can see the events. I can mock what that would look like if we want to look into doing something like that now

Jake3231:
@RubenOlano I made a minor change to make variable naming consistent with `develop` to resolve a merge conflict. Does that code still look like it should?

RubenOlano:
@Jake3231 👍

PR #74, opened 2022-10-26
71 image loading

RubenOlano:
Need someone to test if this solution works better. Should be a bit faster with the priority tag


Jake3231:
In a quick test I saw better performance, though I'd have to test it more extensively.

Jake3231:
Hopefully eventually we can set up automated testing for some stuff like that

Jake3231:
If we can resolve the Vercel issue, that may be helpful to verify the fix, though if not it may be helpful for multiple people to test if this is speeding up image loading. At least when I briefly tested it seemed to be better.

Jake3231 (dismissed):
Tentatively approving this, as it seems to help and I don't see any big potential downsides to this change, and we can trivially change it if further testing warrants.

RubenOlano:
Looks a lot better to me now

Jake3231 (approved):
LGTM! With the further changes it's even faster!

PR #89, opened 2022-11-02
Update [orgName].tsx

dlaunikitis:
Commented out a comment that was causing the next app to fail the build process


RubenOlano (approved):
lgtm

Jake3231:
Looks like this is a good change, though we will want to investigate how to add this behavior back.

RubenOlano:
I think one way to fix this would be to redirect within a getServerSideProps function rather than checking within the component.

PR #100, opened 2022-11-16
Rewriting with daisyUI

RubenOlano:
Draft for now


RubenOlano:
Might want to talk to the team about a theme we can use for this

Jake3231:
Is this still relevant considering #123 ?

RubenOlano:
I think we can close this one, with the new redesign and upcoming changes we don't really need this.

PR #102, opened 2022-12-06
Remove tags from directory UI

Jake3231:
Per last week's team meeting, tags are going to be removed for the initial release. This PR simply removes the tags from the directory UI, but leaves them as shown in the org detail page to allow further design decisions to be made around the usage of tags there.
Screenshot 2022-12-06 at 12 08 58 PM


Jake3231:
I realize this should probably also remove tags from the filter UI!

RubenOlano (approved):
Looks good 👍

Buchimi (approved):
I approve of removing tags.

I'm also yet to decide how to integrate them. 

Though they are unnecessary for a 1.0 release. 

But we should tackle this problem in a 1.x release

Buchimi:
@Jake3231 

You can make the merge

Jake3231:
> I approve of removing tags.
> 
> I'm also yet to decide how to integrate them.
> 
> Though they are unnecessary for a 1.0 release.
> 
> But we should tackle this problem in a 1.x release

Yeah, It would be great to consider this for a later release!

PR #103, opened 2023-02-03
aligned contact

felixjimenez4:
just changed line to ass 12px of padding


Jake3231 (approved):
Thanks for making that fix!

RubenOlano (approved):
Good change

PR #113, opened 2023-02-13
More dynamic contact buttons

Jake3231:
This PR enables contact buttons in the org page to be more dynamic. For common contact methods (Discord and Instagram right now) it adds a color to the button. This is structured so that we can easily add new styles later.

It also adds support for linking to external pages when clicking a button.
Screenshot 2023-02-13 at 1 32 47 PM

### DIFF
--- a/components/ContactList.tsx
+++ b/components/ContactList.tsx
@@ -10,12 +10,14 @@ const ContactList: FC<ContactListProps> = ({ contactMethods }) => {
   return (
     <>
       {contactMethods.map((contact, i) => (
+        <a href="https://www.utdnebula.com">
         <div
           key={i}
           className="flex flex-row justify-center items-center p-3 m-3 rounded-2xl bg-gray-200 hover:bg-gray-300 cursor-pointer"
           style={{backgroundColor:(contactStyles[contact] ?? contactStyles["Default"]).bg, color:(contactStyles[contact] ?? contactStyles["Default"]).text}}  >
           <p className="text-2xl font-bold">{contact}</p>
         </div>
+        </a>
       ))}
     </>
   );
--- a/components/ContactList.tsx
+++ b/components/ContactList.tsx
@@ -1,5 +1,6 @@
 import React, { FC } from 'react';
 import OrgDirectoryTag from './OrgDirectoryTag';
+import contactStyles from '../contactButtonStyles.json';
 
 interface ContactListProps {
   contactMethods: string[];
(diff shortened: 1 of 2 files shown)


Jake3231:
I'd be interested in feedback on the changes in `ContactList.tsx`. I feel like there's got to be a more optimal way to do what I'm doing.

Buchimi:
Why does the vercel build fail?

Jake3231:
> Why does the vercel build fail?

I'm having some trouble with the React code in [ContactList.tsx](https://github.com/UTDNebula/jupiter/pull/113#diff-2720f052a3464976c18b0efd9f193506f406c2cfbb68da1f1f8b808fcff41203), I'm trying to do something that's a bit hacky but technically does work with a build error. Currently I'm leaving the PR as a draft until I can resolve that.

Jake3231:
@felixjimenez4 Do you know if there's a better way to do this?

RubenOlano:
Attempted some fixes, should build now

Jake3231:
Thanks for making those fixes @RubenOlano! Looks like a better solution than what I was attempting.

contactButtonStyles.json
```diff
@@ -0,0 +1,11 @@
+{
+  "Discord": {
+    "bg": "#7289da",
+    "text": "text-white"
+  },
+  "Instagram": {
+    "bg": "#E138B6",
+    "text": "text-white"
+  },
+  "Email": {}
```
RubenOlano, line 10: Instead of using a json file make it an export from a TS file so that it can be extended with interfaces

components/ContactList.tsx
```diff
@@ -1,21 +1,34 @@
...
     <>
-      {contactMethods.map((contact, i) => (
-        <div
-          key={i}
-          className="flex flex-row justify-center items-center p-3 m-3 rounded-2xl bg-gray-200 hover:bg-gray-300 cursor-pointer"
-        >
-          <p className="text-2xl font-bold">{contact}</p>
-        </div>
-      ))}
+      {contactMethods.map((contact, i) => {
```
RubenOlano, line 13: This was a quick fix, definitely try to find a better solution

RubenOlano (changes requested):
Looks good from what I see just a little messy

contactButtonStyles.json
Jake3231: That's a good idea. The motivation of using JSON was to make it more easily updatable if new contact methods are needed. Maybe it would be just as easy to actually update the source code.

PR #114, opened 2023-02-15
Pulling firebase utilities

Buchimi:
I was able to create tools and connect our backend code to firebase servers. 
If you have access to the firebase project. You need to create an .env.local file to be able to use firebase and an .env file to run the firebase tests.
Reach out to me on instructions.

install the firebase package and mocha
``` 
npm i 
```


Buchimi:
Looks like i need to do a git pull and get the base branch

Jake3231:
Looks great! It's exciting to see the initial Firebase integration working so well!

backend_tools/db_provider.ts
```diff
@@ -0,0 +1,122 @@
...
+        }
+        catch (error) {
+            console.error(`failure saving user, got error: ${error}`);
+            return false;
+        }
+    }
+
+    async createClub(club: Club): Promise<boolean> {
+        try {
+            //const clubCollection = collection(this.db, this.club_path)
```
Jake3231, line 30: Is this line needed currently?

backend_tools/db_provider.ts
```diff
@@ -0,0 +1,122 @@
...
+        return ref.data() as unknown as User;
+    }
+
+    async getClub(club_name: string): Promise<Club> {
+        const clubReference = doc(this.db, this.club_path, club_name)
+        const ref = await getDoc<DocumentData>(clubReference)
+        return ref.data() as Club;
+    }
+
+    async getSomeClubs(): Promise<Club[] | null> {
```
Jake3231, line 55: Could this potentially be renamed to best indicate what it is doing, if I'm understanding correctly? 
```suggestion
    async getAllClubs(): Promise<Club[] | null> {
```

backend_tools/picture_uploader.ts
```diff
@@ -0,0 +1,36 @@
+import { deleteObject, getBlob, getStorage, ref, uploadBytes, UploadResult } from "firebase/storage"
+class PictureUploader {
+    storage = getStorage()
+    image_path = "images"
+
+    /// Uploads an image to our storage bucket
+    /// if image name is undefined, it
```
Jake3231, line 7: Is this comment fully completed?

backend_tools/test_db_provider.ts
```diff
@@ -0,0 +1,8 @@
+import DbProvider from "./db_provider";
+
+class DbTestProvider extends DbProvider {
+    user_path: string = "TestUsers"
```
Jake3231, line 4: Should these potentially be changed to be production paths?
```suggestion
    user_path: string = "Users"
```

models/club.ts
```diff
@@ -0,0 +1,11 @@
+import { DocumentReference, DocumentData } from "firebase/firestore"
+import Contacts from "./contacts"
+
+interface Club {
+    name: string
+    events?: DocumentReference<DocumentData>[] //references to other events,
```
Jake3231, line 6: Is this comment partially cut off? It may have not come through correctly.

models/contacts.ts
```diff
@@ -0,0 +1,13 @@
+interface Contacts {
```
Jake3231, line 1: This seems great to define a structure, though would we want to implement a solution that could be more general as to account for contact methods we didn't think of?

models/event.ts
```diff
@@ -0,0 +1,8 @@
+interface Event {
+    name: string,
+    hostclub: string    //club name. THIS NEEDS TO BE THE SAME AS THE CLUB ID IN THE DATABASE.
```
Jake3231, line 3: Per the comment, does anything need to change here prior to merging?

tests/test.ts
```diff
@@ -0,0 +1,93 @@
...
+import { config } from "dotenv"
+import Firebase from "../backend_tools/firebase"
+import User from "../models/user"
+import Year from "../models/year"
+import Role from "../models/role"
+import Career from "../models/career"
+import DbTestProvider from "../backend_tools/test_db_provider"
+
+
+// const firebaseConfig = {
```
Jake3231, line 13: If the commented code is completely unneeded, it may be good to remove it

Jake3231 (changes requested):
Overall looks great, just left a few specific questions.

backend_tools/db_provider.ts
Buchimi: No it's not. will be deleted

backend_tools/db_provider.ts
Buchimi: You're right

backend_tools/picture_uploader.ts
Buchimi: No, i completed it

backend_tools/test_db_provider.ts
Buchimi: No they shouldn't
These paths are for testing purposes only.

Our test scripts will make use of the db provider for testing

models/event.ts
Buchimi: We can remove the comment. It looks like the club name is registered as it's ID.

models/contacts.ts
Buchimi: I guess we could have an opportunity to upload a link?

models/club.ts
Buchimi: I may need to revise the club datastructure first

models/event.ts
Buchimi: Sorry, this is an event. Not a club. 
I don't think we should make use of functions that work with events yet

tests/test.ts
Buchimi: i removed it

Buchimi:
I have made the requested changes

Jake3231 (commented):
Overall looks good!

pages/index.tsx
```diff
@@ -2,10 +2,9 @@ import Head from 'next/head';
 import Image from 'next/image';
 import EventCard from '../components/EventCard';
 import banner from '../public/banner.png';
-
```
Jake3231, line 5: Is it expected that this space is removed?

Jake3231 (approved):

pages/index.tsx
Buchimi: It's just a space, shouldn't matter

models/club.ts
```diff
@@ -0,0 +1,12 @@
+import { DocumentReference, DocumentData } from "firebase/firestore"
+import Contacts from "./contacts"
+import Event from "./event"
+
+interface Club {
```
ZzRanger, line 5: Just a suggestion...maybe put all your models in the same file or colocate your models with your query methods. A lot of these model files are very small so might be easier to condense it in a single place.

ZzRanger (approved):
Overall looks good! From my experience I've noticed that I end up changing my data models and query methods a lot when building out an MVP. Consider potentially finding ways to colocate information so that the burden of changing the models or the query methods is less taxing.

backend_tools/db_provider.ts
```diff
@@ -0,0 +1,125 @@
...
+      const userCollection = collection(this.db, this.user_path);
+      await addDoc(userCollection, user);
+      return true;
+    } catch (error) {
+      console.error(`failure saving user, got error: ${error}`);
+      return false;
+    }
+  }
+
+  async createClub(club: Club): Promise<boolean> {
```
ZzRanger, line 41: Maybe instead of returning true or false return the club id | undefined

ZzRanger: Same thing w/ user

backend_tools/db_provider.ts
```diff
@@ -0,0 +1,125 @@
...
+      return true;
+    } catch (error) {
+      console.error(`failure saving club, got error: ${error}`);
+      return false;
+    }
+  }
+
+  async getUser(userid: string): Promise<User> {
+    const userReference = doc(this.db, this.user_path, userid);
+    const ref = await getDoc<DocumentData>(userReference);
```
ZzRanger, line 56: How does firebase handle if a resource doesn't exist given a certain query? Might be nice to potentially add some error state handling for that here

backend_tools/db_provider.ts
Buchimi: The club id for this function is the name of the club

backend_tools/db_provider.ts
Buchimi: returns undefined

models/club.ts
Buchimi: i see, i prefer them as multiple files. don't they get optimized away by webpack or some similar technology?

RubenOlano (approved):
I think this looks good, I think we can maybe use a parser library to handle some of the parsings whenever passing data to frontend, but we can look at that in the future

Jake3231 (approved):
These changes look good!

PR #119, opened 2023-03-21
104/directory search

### LINKED ISSUE #104: Enable the search bar in the Directory
# Overview

As a user, I should be able to search for orgs using a search field in the Directory page. This is a critical feature for a basic use case.

Likely, the search functionality should consider the title of each org, and potentially the description. For performance reasons, we should consider what data we want to search.

For this issue, typing into the search field should reduce the number of results shown in the Directory. The search could happen after the user presses the enter key, or after the user has stopped typing for x seconds (likely both should be supported). Autocomplete is not needed at this stage. Additionally, animating the results changing would be nice to have but is not critical for this initial implementation.
# Proposal

- [ ] Implement basic backend support
- [ ] Enable filtering displayed orgs
- [ ] Filter based upon text entered into search bar

RubenOlano:
The first initial run of how the search may work along with some changes made to how the org page is searched. I created an API route that queries the db helper, then when clicking on a club, you get redirected to the specific club. I'm using getStaticPaths and getStaticProps to make sure that the routes match up to what's in the db


RubenOlano:
This kinda broke the directory page since the tags and images aren't in the firestore but we can fix those once we start adding to firestore

Jake3231:
> This kinda broke the directory page since the tags and images aren't in the firestore but we can fix those once we start adding to firestore

We may want to either add a placeholder UI when an image doesn't exist, or at least fix the UI to be correctly aligned without an image specified. This could come up again if we don't require all clubs to supply a logo, or if there's any sort of issue with it.

components/DirectorySearch.tsx
```diff
@@ -0,0 +1,69 @@
...
+import Club from '../models/club';
+
+const fetchResults = async (name: string) => {
+  const res = await fetch(`/api/club?name=${name}`);
+  const data = await res.json();
+  if (!data?.clubs || name === '') return [];
+  return data.clubs;
+};
+
+const DirectorySearch = () => {
```
Jake3231, line 12: Is this component used only on the homepage, or also in the Directory search field? If it's only for the homepage it might make sense to have a more specific name.

RubenOlano:
> We may want to either add a placeholder UI when an image doesn't exist, or at least fix the UI to be correctly aligned without an image specified. This could come up again if we don't require all clubs to supply a logo, or if there's any sort of issue with it.

In the function that parses the data, we can set the defaults there for images so that clubs don't have to instantly set an image and tags

components/DirectorySearch.tsx
RubenOlano: I think you're right because the other search bar will work a little bit different because of the grid

pages/directory/[orgName].tsx
```diff
@@ -34,14 +33,6 @@ const OrganizationPage = () => {
...
-            <Image
-              src={org.imageLink}
-              alt={org.name}
-              width={250}
-              height={250}
-            />
-          </div>
           <div className="md:col-span-2">
             <p>
               Lorem ipsum dolor sit amet consectetur adipisicing elit. Esse
```
Comment (Jake3231) [inline pages/directory/[orgName].tsx:38]: Could this potentially use the org.description field from Firestore also?

Comment (RubenOlano) [reply pages/directory/[orgName].tsx:38]: Do you mean like writing that inside of the p tag?

backend_tools/db_provider.ts
```diff
@@ -80,7 +80,14 @@ class DbProvider {
 
   async getClubsByName(name: string): Promise<Club[]> {
     const clubRef = collection(this.db, this.clubPath);
-    const q: Query<DocumentData> = query(clubRef, where('name', '==', name));
+    const capitalizedName = name.charAt(0).toUpperCase() + name.slice(1);
+
+    const q: Query<DocumentData> = query(
```
Jake3231, line 85: This looks like a great change to support searching with the first character non-capitalized or capitalized. Maybe it would make sense to support searching with any capitalization by performing all searches as if every letter of the name is lowercase. Perhaps we could have an auto-generated "searchName" field in Firebase including a sanitized org name that is easier to query.

One issue I ran into with a quick test is that searching "NEBUL" returns no autocomplete results, while "Nebul" works as expected.

RubenOlano: Yea, I was gonna try to implement a fuzzy search algorithm but I want to discuss with the team to see if it may cause unnecessary queries

components/DirectorySearch.tsx
```diff
@@ -0,0 +1,69 @@
...
+      </p>
+      <div className="relative">
+        <input
+          className="relative mt-5 md:mt-10 rounded-md md:text-2xl w-full h-12 md:h-16 px-4 md:px-8 bg-white"
+          placeholder="Search Organizations"
+          autoFocus
+          tabIndex={1}
+          onChange={handleSearch}
+        />
+        {results.length > 0 && (
```
Jake3231, line 51: Overall this looks great. We should consider if it would make sense to slightly separate the autocomplete results visually from the search field, or some other way to visually integrate it with the search field.

pages/api/club.ts
```diff
@@ -0,0 +1,22 @@
+import { NextApiRequest, NextApiResponse } from 'next';
+import DbProvider from '../../backend_tools/db_provider';
+
+export default async function handler(
```
Jake3231, line 4: The implementation looks good, and it may be helpful to add a brief comment describing what this function does and how it's intended to be used.

pages/directory/[orgName].tsx
```diff
@@ -60,3 +51,36 @@ const OrganizationPage = () => {
 };
 
 export default OrganizationPage;
+
+export const getStaticPaths: GetStaticPaths = async () => {
+  const db = new DbProvider();
+  const clubs = await db.getAllClubs();
```
Comment (Jake3231) [inline pages/directory/[orgName].tsx:57]: Is it correct that this requests a list of all orgs when loading any individual page? It seems like it'll be important to ensure we're caching this endpoint when possible for performance and reducing API calls when reasonable.

Comment (RubenOlano) [reply pages/directory/[orgName].tsx:63]: This function only runs at build time, it ensures that a page exists for all clubs in the database. Caching should not be necessary

pages/directory/[orgName].tsx
```diff
@@ -60,3 +51,36 @@ const OrganizationPage = () => {
...
+  }));
+
+  return { paths, fallback: false };
+};
+
+export const getStaticProps: GetStaticProps<{ club: Club }> = async (ctx) => {
+  const db = new DbProvider();
+  const orgName = ctx?.params?.orgName as string;
+  const clubs = await db.getClubsByName(orgName);
+  const club = clubs[0];
```
Comment (Jake3231) [inline pages/directory/[orgName].tsx:84]: It looks like the code is assuming that the org the user selected will always be the first item in the results list returned from the API call. This is likely true most of the time, though it looks like there could be a case where this isn't true - does this still work in a scenario where we had clubs named "Org 1" and "Org 2" where the user selects "Org 2"?

Jake3231 (commented):
Overall looks pretty good, just left a few questions.

pages/directory/[orgName].tsx
Comment (RubenOlano) [reply pages/directory/[orgName].tsx:84]: I was thinking of asking in the next team meeting if we want to move away from using the org names and instead using the ids that firebase gives us. That would eliminate the need for us to search by name whenever loading the page. Just a thought

pages/directory/[orgName].tsx
Comment (Jake3231) [reply pages/directory/[orgName].tsx:63]: Thanks for clarifying that! I had a misconception about this, and have re-read the docs to make sure I understood this correctly.

pages/directory/[orgName].tsx
Comment (RubenOlano) [reply pages/directory/[orgName].tsx:84]: Also, this issue may not arise because when we load the page, we are passing in the full name of the organization that the user chose, unless a club has an identical name, then there shouldn't be any issues

Comment (Jake3231) [reply pages/directory/[orgName].tsx:84]: In general, I think it would be a good idea to support loading a club by ID. To be sure I've got it, this implementation will return the correct org based on matching the entire name?

Comment (RubenOlano) [reply pages/directory/[orgName].tsx:84]: Yes

Comment (Jake3231) [reply pages/directory/[orgName].tsx:84]: Sounds good. I'll resolve this for now, though we should make support for fetching clubs by ID a relatively new to-do

Buchimi:
To fetch clubs by ID, you can use the name. 
You can also build a query. I have a test on my local computer that proves queries work (for clubs at least)

Jake3231:
> In the function that parses the data, we can set the defaults there for images so that clubs don't have to instantly set an image and tags

That sounds like a pretty good approach. If we're on the same page then the function always results some image to the UI. What would you think of something like a gray square, or other non-visual placeholder as the default?

Jake3231:
> To fetch clubs by ID, you can use the name.

Do you mean that you can pass the ID into the name query?

Buchimi:
Clubs are stored with their names. 
I believe there's a function called "get club by id" you can use that

Buchimi:
For example, to get ACM, you pass ACM as the ID

Jake3231:
I see what you're saying. I think we'd also want a method to fetch clubs using the Firebase record ID, in addition to the club name.

Buchimi:
Doesn't the ID equal the name of the club

RubenOlano:
See if that works better. Update to use ids and should now work for all clubs. Prev implementation failed for the nebula labs test and the ACM club

Buchimi:
Sure but what do you want to store as the club ID? It's pretty random when you make the document of firebase

RubenOlano:
Random key is fine I think

backend_tools/db_provider.ts
RubenOlano: I asked some other people and they recommended Fuse.js so I'll look into that

RubenOlano:
With these changes, we may need to use something like react query to cache the info that is given so we don't constantly send and receive the same info

Jake3231:
The design updates are looking really nice!

Jake3231:
> With these changes, we may need to use something like react query to cache the info that is given so we don't constantly send and receive the same info

That sounds like a great idea. It may be good to split that into another branch/issue, unless that would be an easy change. I suspect properly caching org data wouldn't be trivial.

RubenOlano:
> That sounds like a great idea. It may be good to split that into another branch/issue, unless that would be an easy change. I suspect properly caching org data wouldn't be trivial.

I'll see how easy/difficult this is and integrate it to this if its not too difficult

RubenOlano:
#121 implements this

PR #120, opened 2023-03-22
User settings account

### LINKED ISSUE #106: User account support needed
# Overview

As a user or an administrator, I should be able to access Jupiter with varying permissions enabled.

Once we have basic backend integration working, it'll be necessary to establish a structure and functionality for user accounts. This way, we can support saving data and allow certain users permission to edit org data.

# Proposal

- [ ] Establish account roles (ex: User, Org owner, Org rep, admin, etc)
- [ ] Implement user roles in the backend
- [ ] Wire up UI visibility and access according to user role

RubenOlano:
Closes #106 
 This is the first iteration of the user settings account. I'm using next auth to simplify authentication and connection to the firestore database. Whenever a user initially signs in, a new account is created and registered to their ID, this allows us to access their data and update within the settings page. Added a type declaration file in order to ease the access of the data without needing to do a bunch of checks. Let me know if you encounter any errors or have any ideas on how to make this cleaner / less buggy


RubenOlano:
I added some stuff to the README. When testing you need to get service account credentials and input that into the .env file

Jake3231:
> I added some stuff to the README. When testing you need to get service account credentials and input that into the .env file

Thanks for adding that info! We should be able to update our deployment as needed.

Jake3231:
cc @shreyon312

RubenOlano:
I may close this branch in favor of the TRPC branch just because It's already implemented better In that branch

RubenOlano:
We can also redesign with the new implementation we are using

Issue #107, opened 2023-02-10
New `Event` page needed


Jake3231:
# Overview

As a user, I should be able to select an event and view detailed event information. This Issue is tracking work to bring up a new `Event` page that can show detailed information about an org's event.

Likely, this should include some sort of image, title, description, date/time, interested participants count, and possibly more.

# Proposal

- [ ] Create new Event page that is accessed by clicking an event in the Events browser
- [ ] Create draft layout for key information
- [ ] Allow the page to be populated with non-placeholder data when sample events are clicked.

Master7130:
Can we rename the events directory page to be called "eventsDirectory" from "events." This would reduce confusion between the page I'm going to be creating, "event." Or we could call this page something else.

Jake3231:
@Master7130 I may have missed a previous comment, were you saying that the naming of the page might be unclear?

Master7130:
Yeah that was my initial thought and then I realized I could just make a folder containing the events page and the dynamic route for each event like for the directory.

Master7130:
Someone seems to have created an events page already. What would be the next steps for this? Pull data from firebase to populate it instead of dummy data?

RubenOlano:
I think that’d be a good idea. We can also extend events to include
descriptions and have a page for individual events

Master7130:
Is this good to be closed?

PR #122, opened 2023-04-01
Events Page

Master7130:
Created a dynamic route for an event details page. Can be accessed by clicking on the title of the event on the events page. 

#107

### DIFF
--- a/components/Event.tsx
+++ b/components/Event.tsx
@@ -1,6 +1,7 @@
 import React, { FC, useState } from 'react';
 import demoEvents from '../demoEvents.json';
 import Image from 'next/image';
+import Link from 'next/link';
 
 interface Props {
   event: typeof demoEvents[0];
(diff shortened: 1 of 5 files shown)


Jake3231:
We should consider how this also could fit with #123

Buchimi:
There's a way to do this with nextjs routing system. 

I imagine we'll need the backend to provide the data we need to create the page

Jake3231:
@Master7130 Thanks again for making this change! Per the recent meetings, as it sounds like we'll be moving forward with the redesign branch, do you think this could be easily ported to that as well? We could merge a design into develop now, and add it back to the redesign later also.

Master7130:
Yeah it shouldn't be too difficult to have something like this on the redesign branch. Would you like me to do it or will someone else handle it?

Jake3231:
> Yeah it shouldn't be too difficult to have something like this on the redesign branch. Would you like me to do it or will someone else handle it?

If you are able to add that, that would be awesome. I'd suggest branching from `redesign`, and also checking with Solomon and Ruben regarding design and implementation details

PR #129, opened 2023-09-19
next/router -> next/navigation

Master7130:
Moved from next/router (next 12) to next/navigation (next 13) and made small modifications based on api changes.


RubenOlano:
Are we good to use `next/navigation` in the pages folder? I know in the documentation it says to upgrade to it when migrating to the app router but since we're still using pages is it chill or should we expect some different behavior? For now it seems like its working fine but I just want to make sure that we're not going to break something in the future

Issue #130, opened 2023-09-19
Sign In Page


RubenOlano:
# Overview

As a user, I should be able to use a login page in order to authenticate.

For now, we don't have a dedicated platform to sign in so this will be based on the current [planner](https://planner.utdnebula.com/auth/login) sign-in flow

<img width="840" alt="image" src="https://github.com/UTDNebula/jupiter/assets/76502130/52b32ae0-6516-4c6c-9ea6-6ae2a6c3159b">
For the time being, try to keep it limited to using Google as the provider

Make sure you understand how next-auth works with redirects in order to properly implement the flow and make a seamless experience for the user.

Master7130:
Can I work on this?

RubenOlano:
I can assign it to you for now but I believe Daniel is working on implementing NextAuth at the moment. So for now try implementing the style and look and once Daniel's done you can try integrating it to the flow

PR #134, opened 2023-09-19
Sign in and sign up pages

Master7130:
Sign in and sign up pages based on Planner's implementation.

#130


src/pages/auth.tsx
```diff
@@ -0,0 +1,68 @@
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
```
RubenOlano, line 1: We should be able to remove these lines, all that needs to happen is to add the keyword `void` on promises

src/pages/auth.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const toggleAuthType = () => {
+    setSignin(signIn => !signIn)
+  }
+
+  return (
+    <main className="relative flex h-screen flex-col items-center justify-center space-y-10 bg-[#ffffff] md:pl-72">
+      <h1 className="text-2xl text-slate-800">{signin ? "Sign in" : "Sign up"}</h1>
+      {Object.values(providers).map((provider) => (
+        <div key={provider.name}>
+          <button onClick={() => signIn(provider.id)}>
```
RubenOlano, line 26: `() => void signIn(provider.id)`

RubenOlano (changes requested):

RubenOlano (approved):

PR #135, opened 2023-09-21
124 user authentication

danielsuhcompsci:
Just added the adapter and changed user -> userMetadata where needed


RubenOlano (approved):
Looks good, `User` has been changed to `UserMetadata` for clarity and I removed some unnecessary env variables as well. Tests failing is weird since nothing fundamentally changed but we can look at that in a different pr

Master7130 (approved):

PR #136, opened 2023-09-24
dynamic-buttons-contact

MistahSanta:
Note: This is my first pull request, so feel free to be as honest as possible, so I can learn  whether that is in my naming, coding logic, etc

Feature: 
- Creates the correct number of buttons to each UTD organization social media platform. I.E. let say UTD Nebula has 4 social media platform, so when the user clicks on the "learn more" four buttons should appear that link to each social media platform UTD nebula uses. 

My approach: 
1) Change the Contact object to accept the different types of platform as well as the url to the social media platform 
2) Create a dummy data set to test out the contact button 
3) Create a tsx file that will load the Contact Buttons
- This file will map to each element in the Contact array and connect the correct Icon and link to that button

### DIFF
--- a/src/components/ContactButtons.tsx
+++ b/src/components/ContactButtons.tsx
@@ -1,0 +1,53 @@
+import { type Contacts } from '@src/models/contacts';
+import { FC } from 'react';
+import Image from 'next/image';
+
+type logoProps = {
+  discord: string;
+  youtube: string;
+  twitch: string;
+  facebook: string;
+  twitter: string;
+  instagram: string;
+  website: string;
+  email: string;
+  other: string;
+};
+
+const logo: logoProps = {
+  discord: '/favicon-16x16.png',
+  youtube: '/nebula-logo.png',
+  twitch: '/nebula-logo.png',
+  facebook: '/nebula-logo.png',
+  twitter: '/nebula-logo.png',
+  instagram: '/nebula-logo.png',
+  website: '/nebula-logo.png',
+  email: '/nebula-logo.png',
+  other: '/Jupiter.png',
+};
+
+const ContactButtons: FC<{ contact: Contacts[] }> = ({ contact }) => {
+  return (
+    <>
+      {contact.map((item, key) => (
+        <button
+          key={key}
+          className="relative mx-2 my-5 rounded-full bg-slate-100 px-4 py-1 transition-colors hover:bg-slate-300"
+        >
+          <a target="__blank__" href={item.url}>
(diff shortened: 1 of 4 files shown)


src/components/ContactButtons.tsx
```diff
@@ -0,0 +1,53 @@
...
+  email: '/nebula-logo.png',
+  other: '/Jupiter.png',
+};
+
+const ContactButtons: FC<{ contact: Contacts[] }> = ({ contact }) => {
+  return (
+    <>
+      {contact.map((item, key) => (
+        <button
+          key={key}
```
RubenOlano, line 34: Instead of using the index as a key, use the URL since it should be unique for every item

RubenOlano (changes requested):
Make sure to also update the IClub object so it takes in an array of the contact type

src/components/ContactButtons.tsx
MistahSanta: Updated the key to accept url as well as Iclub object to take an array of Contact

src/components/OrgHeader.tsx
```diff
@@ -30,6 +53,7 @@ const OrgHeader: FC<{ club: Club }> = ({ club }) => {
         <button className="m-5 rounded-full bg-slate-100 px-4 py-2 font-semibold text-slate-900 transition-colors hover:bg-slate-300">
           Join
         </button>
+        <ContactButtons contact={ dummyData || []}/>
```
RubenOlano, line 56: now that it's updated try passing in the contacts based on the club object since now it is guaranteed to default to an empty array

RubenOlano (changes requested):

src/components/OrgHeader.tsx
MistahSanta: Update: Passed club.contract and successfully loaded one button that connects to "google.com." Also removed dummy data set.

RubenOlano: I think the google.com is just dummy data I added to the db.

RubenOlano (approved):
Looks good 👍🏽

Master7130:
Are we worrying about responsivity or are new issues going to be created to add breakpoints to existing pages/components?

image

RubenOlano:
> Are we worrying about responsivity or are new issues going to be created to add breakpoints to existing pages/components?
> 
> image

Yeah, that comes in the future when we get a mobile design for some of these pages. We also might need to have a discussion about a limit on how many of the contact options show up on the front page at once

Master7130 (approved):

PR #138, opened 2023-09-25
Migrate db

RubenOlano:
Moving towards Postgres to lighten the dependencies needed for Jupiter. Should retain the same functionality, will just need to update the tests in order to properly reflect that


src/pages/directory/[clubId].tsx
```diff
@@ -34,16 +34,14 @@ import {
...
 import { appRouter } from '@src/server/api/root';
 import { createInnerTRPCContext } from '@src/server/api/trpc';
 import { api } from '@src/utils/api';
+import { db } from '@src/server/db';
 
 export const getStaticPaths: GetStaticPaths = async () => {
-  const db = new DbProvider();
-  const clubs = await db.getAllClubs();
-
+  const clubs = await db.query.club.findMany();
```
Comment (Master7130) [inline src/pages/directory/[clubId].tsx:44]: Is there a reason you're making a direct call to the db instead of through trpc (`const clubs = api.club.all()`)?

Comment (RubenOlano) [reply src/pages/directory/[clubId].tsx:44]: The trpc `api` object is for use within react components. `getStaticPaths` happens at compile time so that's why we're using it that way

Comment (Master7130) [reply src/pages/directory/[clubId].tsx:44]: 👍

Comment (Master7130) [inline src/pages/directory/[clubId].tsx:44]: Should `{ with: { contacts: true } }` be added like it is in the trpc definition to get all clubs? Why is this condition here anyway? Are we not going to show clubs that have no contact info?

src/components/ContactButtons.tsx
```diff
@@ -1,5 +1,5 @@
 import { type Contacts } from '@src/models/contacts';
-import { FC } from 'react';
+import { type FC } from 'react';
```
Master7130, line 2: `import type { FC } from 'react';`

I think we should import types using the above syntax so that if we are ever importing multiple types, we don't have to keep writing 'type' in front of each one. It would also make all of our code more consistent.

src/server/api/routers/club.ts
```diff
@@ -12,27 +15,46 @@ const byIdSchema = z.object({
...
-      return await ctx.dbProvider.getClubById(id);
+      const byId = await ctx.db.query.club.findFirst({
+        where: (club) => eq(club.id, id),
+        with: { contacts: true },
+      });
+
+      const first = IClub.parse(byId);
+      return first;
     } catch (e) {
       console.log(e);
```
Master7130, line 44: Minor but to keep consistent with line 57, it should be `console.error()` here as well.

src/server/db/schema.ts
```diff
@@ -0,0 +1,103 @@
...
+  'website',
+  'email',
+  'other',
+]);
+
+export const contacts = pgTable(
+  'contacts',
+  {
+    platform: platformEnum('platform').notNull(),
+    url: text('url'),
```
Master7130, line 87: Shouldn't url also not be null? There's no point in having a contact shown if a user can't access it.

src/pages/directory/[clubId].tsx
Comment (RubenOlano) [reply src/pages/directory/[clubId].tsx:44]: I added it just so that it matched the previous firebase implementation. One thing that might be beneficial is moving the models away from the `models/` folder and trying to use `drizzle-zod` to generate the types based on the actual schema we create

Comment (Master7130) [reply src/pages/directory/[clubId].tsx:44]: Yeah, that would remove a lot of the work we would have to do. Everything else looks good. I'll approve the pull request and you can merge when you're ready.

Master7130 (dismissed):

src/components/OrgHeader.tsx
```diff
@@ -1,7 +1,14 @@
 import React, { type FC } from 'react';
 import Image from 'next/image';
-import { type Club } from '@src/models/club';
 import ContactButtons from './ContactButtons';
+import type {
+  SelectClub,
```
Master7130, line 5: Should probably be import as `Club` like it's being done in other files.

src/server/api/routers/club.ts
```diff
@@ -1,6 +1,8 @@
+import { eq } from 'drizzle-orm';
 import { createTRPCRouter, publicProcedure } from '../trpc';
 import { z } from 'zod';
-
+import Fuse from 'fuse.js';
+import { selectClub } from '@src/server/db/models';
```
Master7130, line 5: `import { selectClub } from '@src/server/db/models';` -> `import { selectClub as Club } from '@src/server/db/models';`

Issue #137, opened 2023-09-25
Search bar disappears


RubenOlano:
### What did you expect to happen?

I expected to be able to search for orgs within an org info page

### What happened?

Currently the search bar disappears whenever navigating to the info page 

### Version

Pre-release (Development)

### What browsers are you seeing the problem on?

Chrome

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

KennethImphean:
Could I work on this even though I wasn't matched to the project?

RubenOlano:
Go for it, just make a separate branch and make a pull request once you're ready for code review

PR #140, opened 2023-09-25
132 padding and styling fix

### LINKED ISSUE #132: Padding Issues
# Overview

As a user, I should be able to see the correct design for the Jupiter project

Currently, the implemented design only really follows the design at a distance. A lot of the margins and padding are not ideal and require some tweaking

Look at the [Jupiter Figma](https://www.figma.com/file/snmEoRUgfjAZYSc6hqySkY/Wireframes-for-Jupiter?type=design&node-id=1329%3A1637&mode=design&t=BmEIiTjucdjdHbb8-1) (under pages that have `redesign`) to look at how padding and margins should properly be implemented

Use the dev feature for Figma as it makes it easier to implement

nl32:
Updates the following to better match the design spec in Figma.
- Homepage
- About
- Club info
- Sidebar
---
Fixes #132 and #137


RubenOlano (changes requested):
Looks much better now! I don't know if this may be something with the design, but the search bar for events and the search bar for clubs seems to be slightly off whenever changing between the homepage and the events tab. Other than that, some of the button text color seems to be a bit off so try to see if you can fix take a peek at that. But overall, good job so far with this, and even tackling two issues for the price of one is great

nl32:
I fixed the search bars looking different and it should be reusable now if we need a search bar elsewhere in the app. For the button colors I fixed the blues and a few texts being the wrong color.

RubenOlano (dismissed):
One last little nitpick, the body of the About page seems to not take up the entire width of the screen. I'll approve for now, but once you fix that, ping so we can merge it in

nl32:
@RubenOlano I fixed the about page cutting off and I added a like button and footer to the club page.

RubenOlano (approved):
LGTM 👍🏽

PR #145, opened 2023-09-27
Move to jest for typescript compatibility

RubenOlano:
Mocha/chai was not working super well so I tried moving to jest. Might need to consider making separate tables for testing or a separate project for tests


tests/test.ts
```diff
@@ -1,37 +1,31 @@
...
-      .returning()
-      .then((res) => {
-        assert(res.length > 0);
-        assert.equal(res[0]?.name, 'TEST ORG');
-        done();
-      })
-      .catch(done);
-  }).timeout(5000);
+    const returned = await db.insert(club).values(newCLub).returning();
+    expect(returned.length > 0).toBeTruthy();
```
Master7130, line 15: I'm assuming `returned` should contain only the entry that was inserted into the db. If that's the case I think it would be best to modify the condition to `returned.length == 0`

RubenOlano: Fixed to just be `returned.length === 1`

Master7130 (approved):

PR #146, opened 2023-09-28
Events page

Master7130:
#107

Created events model, router and some procedures to get events. Events are being fetched from the db and and are shown under the correct date. A modal (radixui dialog) is now being displayed when an event is clicked to show more details.


src/server/db/schema.ts
```diff
@@ -68,6 +68,19 @@ export const club = pgTable('club', {
...
+  id: integer('id')
+    .default(sql`nanoid(20)`)
+    .primaryKey(),
+  clubId: text('club_id')
+    .notNull()
+    .references(() => club.id, { onDelete: 'cascade' }),
+  name: text('name').notNull(),
+  description: text('description').notNull(),
+  startTime: timestamp('start_time').notNull(),
+  endTime: timestamp('end_time').notNull(),
```
RubenOlano, line 81: Try using this with the timezone included just to make sure that nothing is messed up when different timezones are involved. I know right now its not a huge issue but it could become a bigger issue if we don't change it now

src/types/event.ts
RubenOlano, line 1: Why are we using this on top of the type in the schema file?

src/server/api/routers/event.ts
```diff
@@ -0,0 +1,54 @@
+import { eq, gte, lte, and } from 'drizzle-orm';
+import { createTRPCRouter, publicProcedure } from '../trpc';
+import { z } from 'zod';
+import { selectEvent } from '@src/server/db/models';
+
+const byClubIdSchema = z.object({
+  clubId: z.string().default(''),
+  startTime: z.date(),
+  endTime: z.date(),
+});
```
RubenOlano, line 10: Do we plan on using the same schema for both for convenience or is this just temporary?

RubenOlano (changes requested):
Overall the rough draft looks good, a couple of questions and then just the timestamp thing

Master7130:
Thanks for catching those. Let me know if there's anything else.

RubenOlano (approved):
Looks good 👍🏽

Master7130:
We can merge.

PR #147, opened 2023-09-28
update instructions for setting up a database

nl32:
fixes instructions for setting up a database, matching them with the new migration scripts.


RubenOlano (approved):
👍🏽

PR #148, opened 2023-09-28
Search bar changes

nl32:
In #140 I made changes to how the search bar worked, I've decided this should  be a different pr since It's a more technical change to how search bars are implemented, rather than a styling change, and should be considered independent of the style changes
- header changes
    - I made a BaseHeader, that sets the spacing and layout for the headers but lets the contents be different. then the default header and event header can pass in what they should have inside, In this case the different search bars.
    - This change should keep header padding uniform across the project, but might end up not the best idea if we don't want headers coupled like this.
- search bar component
    - There is a new search bar component that handles how all search bars should look and behave. But a parent component would handle searching.
        - This unifies how search bars look across the app and makes it easier to make new search bars for different things.
        - then we have components to search for specific things(i.e. clubs, events, etc) that can just be plugged into the ui.

### DIFF
--- a/src/components/SearchBar.tsx
+++ b/src/components/SearchBar.tsx
@@ -5,6 +5,9 @@ import React, {
   type ChangeEvent,
 } from 'react';
 import { SearchIcon } from './Icons';
+import { useRouter } from 'next/router';
+import type { SelectClub as Club } from '@src/server/db/models';
+import { api } from '@src/utils/api';
 
 type SearchElement = {
   id: string;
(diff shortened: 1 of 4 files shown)


RubenOlano:
To be clear this closes #137 right?

RubenOlano (approved):
Looks good regardless

nl32:
moved the #137 fix in

RubenOlano (approved):

src/components/SearchBar.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const [res, setRes] = useState<Club[]>([]);
+  api.club.byName.useQuery(
+    { name: search },
+    {
+      onSuccess: (data) => {
+        setTimeout(() => setRes(data), 300);
+        return data;
+      },
+    },
+  );
```
Master7130, line 79: This call is being made when the page loads. This is fine for now but when there are more clubs in the database, page loads are going to get slower. 

Can you run the api call when the search bar value changes? You would need to set the `enabled` field to `false` by default and `onChange` of the search bar value, set the `enabled` state to `true`. Inside `onSuccess` you would be setting the `enabled` state back to false so it doesn't keep running.

Master7130 (changes requested):

src/components/SearchBar.tsx
RubenOlano: True, though this might be out of the scope of the current aim for this pr. We can make a new issue to reduce the amount of API calls made whenever typing

Master7130: Alright

Master7130 (approved):

PR #150, opened 2023-09-29
add debounce and lazy loading to search bar

### LINKED ISSUE #149: Search bar debounce
# Overview

Currently, the way we are searching for orgs in using the search bar will cause the search time to increase as the number of orgs in our database increases. We need to rework the debounce that is used when typing in the search bar

nl32:
Search bar now doesn't query until a search is entered, meaning it won't affect page load times. The search also debounces so that it doesn't fetch as much while the user is typing. this fixes #149

### DIFF
--- a/src/components/SearchBar.tsx
+++ b/src/components/SearchBar.tsx
@@ -3,6 +3,7 @@ import React, {
   type SetStateAction,
   useState,
   type ChangeEvent,
+  useEffect,
 } from 'react';
 import { SearchIcon } from './Icons';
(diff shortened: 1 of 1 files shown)


RubenOlano (approved):
LGTM

RubenOlano:
One thing we can also probably consider is disabling refetch on window refocus just cause we're fetching all the clubs. We also have a console.log somewhere

Master7130 (approved):

Issue #109, opened 2023-02-10
Auth UI needed


Jake3231:
# Overview

As a user, I should be able to sign in to Jupiter using one of a few standard login methods.

Currently, there is no auth UI, though we should create a UI that interfaces with Firebase to allow users to log in with either a username and password, or a Google account.

# Proposal
This likely depends on #99 

For this Issue, only a placeholder UI is needed that can be later wired up to the actual auth code.

- [ ] Create new Auth sheet or page
- [ ] Allow it to be shown/dismissed using sign in button in navbar

Jake3231:
Onboarding is covered by @shreyon312 's UI in #10

RubenOlano:
This was closed by #134

PR #152, opened 2023-10-03
Replaced the place holder profile circle with actual profile image and added a dropdown menu

Master7130:
#109 

I have replaced the placeholder profile icon we have with the user's profile pic from the session. I have also added the sign in and sign out links when the user isn't logged in and a log out link under the dropdown menu.

### DIFF
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -54,7 +54,7 @@ const ProfileDropDown = ({ image }: { image: string }) => {
             className="p-2 hover:cursor-pointer"
             onClick={() => void signOut()}
           >
-            <button>Signout</button>
+            <button>Sign out</button>
           </DropdownMenu.Item>
         </DropdownMenu.Content>
       </DropdownMenu.Portal>
--- a/src/components/Header.tsx
+++ b/src/components/Header.tsx
@@ -1,5 +1,6 @@
 /* eslint-disable react/jsx-no-undef */
-import { type ReactNode } from 'react';
+import type { ReactNode } from 'react';
+import { useState } from 'react';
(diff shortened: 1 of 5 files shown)


src/components/Header.tsx
```diff
@@ -1,17 +1,67 @@
...
+          <div className="h-10 w-10 rounded-full bg-gray-300" />
+        )}
+      </DropdownMenu.Trigger>
+      <DropdownMenu.Portal>
+        <DropdownMenu.Content className="mt-2 w-44 rounded-md border-2 border-black bg-slate-200 text-center">
+          <DropdownMenu.Item
+            className="p-2 hover:cursor-pointer"
+            onClick={() => void signOut()}
+          >
+            <button>Signout</button>
```
RubenOlano, line 57: Sign out should be two words here

src/pages/auth.tsx
```diff
@@ -4,15 +4,25 @@ import { getServerSession } from 'next-auth/next';
...
 import { getProviders } from 'next-auth/react';
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
+import { useRouter } from 'next/router';
 
 export default function Auth({ providers }: { providers: Providers }) {
+  const router = useRouter();
   const [signin, setSignin] = useState(true);
 
+  console.log(router.pathname)
```
RubenOlano, line 14: Remove before merging in

RubenOlano (changes requested):
Other than a couple of nitpicks looks good. Won't worry too much about the exact look till we get design to look at it

RubenOlano (approved):
Looks good 👍🏽

PR #159, opened 2023-10-12
Jup 5 display upcoming events

MistahSanta:
Jira Issue #5 - Link and display upcoming events to organization page 

Note - I did not worry about the design of the displaying the upcoming events - work with design to create a better design 

My Approach: 
- Create a seprate tsx file display the upcoming events in the organization page that pulls from event api and map and display those events
- Inside event api, add two optional parameter that allow for filtering and sorting query: { sortByDate, currentTime ( used as a baseline to filter and find events that are in the future ) 
- Edit the event query to be more dynamic that check for the two optional parameters and query based on those parameters


RubenOlano:
Try running `npm run build` on your side and fixing the issues you see. See if that makes the preview work

src/components/OrgUpcomingEvents.tsx
```diff
@@ -0,0 +1,32 @@
...
+      <div className="w-full rounded-lg bg-slate-100 p-10">
+        <h1 className="font-medium text-2xl mt-5">
+          Upcoming Events
+        </h1>
+        <div className="flex flex-col items-start justify-evenly  md:flex-row mt-5">
+            {query_events?.map( (event) => (
+                <div key={event.id} className=" bg-blue-500 p-4 rounded-lg cursor-pointer" >
+                  <div className="font-medium">{event.name}</div>
+                  <div>- {event.description}</div>
+                  <div>- {`${event.startTime.getMonth() + 1}/${event.startTime.getDay()}`}</div>
```
RubenOlano, line 21: Should be `getDate` I think. I'm currently 10/0 for one of the months

src/components/OrgUpcomingEvents.tsx
```diff
@@ -0,0 +1,32 @@
...
+  
+    const query_events = (api.event.byClubId.useQuery({ clubId: club.id, currentTime: cur_time, sortByDate: true  })).data
+    
+    return (
+      <div className="w-full rounded-lg bg-slate-100 p-10">
+        <h1 className="font-medium text-2xl mt-5">
+          Upcoming Events
+        </h1>
+        <div className="flex flex-col items-start justify-evenly  md:flex-row mt-5">
+            {query_events?.map( (event) => (
```
RubenOlano, line 17: Try adding some alternate text whenever there are no items in the array

RubenOlano (changes requested):
Changes look good so far

src/components/OrgUpcomingEvents.tsx
MistahSanta: Here is the placeholder text if there is no item the array: 
image

src/components/OrgUpcomingEvents.tsx
```diff
@@ -0,0 +1,40 @@
...
+            ?
+             query_events.map( (event) => (
+                <div key={event.id} className=" bg-blue-500 p-4 rounded-lg cursor-pointer" >
+                  <div className="font-medium">{event.name}</div>
+                  <div>- {event.description}</div>
+                  <div>- {`${event.startTime.getMonth() + 1}/${event.startTime.getDate()}`}</div>
+                </div>
+            ))
+            :
+             <div className="font-medium">There is no upcoming events</div>
```
RubenOlano, line 30: There "are" no upcoming events

RubenOlano (changes requested):
Tiny change

RubenOlano (dismissed):

src/components/OrgUpcomingEvents.tsx
```diff
@@ -0,0 +1,40 @@
+import { api } from "@src/utils/api";
+import { type FC } from "react";
+import { SelectClub as Club } from "@src/server/db/models";
+
+const cur_time = new Date();
+
+const OrgUpcomingEvents:FC<{club_id : string}> = ({club_id}) => {
+
+  const datt = new Date("2025, 5, 4")
```
danielsuhcompsci, line 9: Maybe remove this

MistahSanta: oh good catch! I forgot to remove that. I just removed it

RubenOlano (dismissed):

src/components/OrgUpcomingEvents.tsx
```diff
@@ -0,0 +1,39 @@
+import { api } from "@src/utils/api";
+import { type FC } from "react";
+import { SelectClub as Club } from "@src/server/db/models";
+
+const cur_time = new Date();
```
Master7130, line 5: Move this inside the component. Only the code inside the component will be executed again when the page is refreshed.

Master7130 (changes requested):

src/components/OrgUpcomingEvents.tsx
MistahSanta: Putting the date inside the component causes the date to be refresh too quickly that useQuery does not pull for the event api. This is the console.log(cur_event). It refreshes really quickly.
image

I thought about using useEffect and useState to set the interval to refresh and pull from the API every 60 seconds, but the issue is every minute, the event is reloaded which means the events disappear and reappear again. To fix this, I useState() and set that time whenever the user loads the components. Let me know if there is a better solution than this approach.

danielsuhcompsci (approved):
Only thing I can nit-pick would be just directly de-structuring the query_events into { data }, in case like isLoading is used in the future.

src/components/OrgUpcomingEvents.tsx
```diff
@@ -0,0 +1,38 @@
+import { api } from "@src/utils/api";
+import { type FC } from "react";
+import { useState} from "react";
+
+const OrgUpcomingEvents:FC<{club_id : string}> = ({club_id}) => {
+
+    const [cur_time] = useState( new Date() );
+
+    const query_events = (api.event.byClubId.useQuery({ clubId: club_id, currentTime: cur_time, sortByDate: true  })).data
```
RubenOlano, line 9: It would be a lot better if you could restructure the object rather than wrapping it in parens like that and then accessing data

RubenOlano (changes requested):

src/components/OrgUpcomingEvents.tsx
MistahSanta: I have move those parameter to an object

RubenOlano (approved):

PR #160, opened 2023-10-13
JUP-16 Fix signup issue

Master7130:


RubenOlano:
Worked with a combination of updating deps and removing some stuff.

Master7130 (commented):
Looks good

PR #164, opened 2023-10-19
App router upgrade

RubenOlano:
Just for a quick look, prob won't work since we need node v21 but it's nice to look at :).


nl32 (dismissed):
lgtm 🤩

PR #166, opened 2023-10-23
Added hover state to sidebar

CHarris0812:
Added a hover state so that when you mouse over one of the sidebar items it looks the same as the active item. Also this is my first pull request here so lmk if I'm doing anything wrong.

### DIFF
--- a/src/components/SidebarItems.tsx
+++ b/src/components/SidebarItems.tsx
@@ -1,4 +1,4 @@
-import React, { type FC } from 'react';
+import React, { type FC, useState } from 'react';
 import {
   AboutIcon,
   EventsIcon,
@@ -48,6 +48,16 @@ const SidebarItems: FC<{ cat: union[number]; active: boolean }> = ({
   const router = useRouter();
   const route = routeMap[cat];
 
+  const [mouseOver, setMouseOver] = useState(0);
+  const handleMouseOver = () => {
+    setMouseOver(1);
+    console.log(cat);
+  };
+
+  const handleMouseOut = () => {
+    setMouseOver(0);
+  };
+
   // This should never happen
   // Just so  TS doesn't complain
   if (!route) return null;
@@ -55,8 +65,16 @@ const SidebarItems: FC<{ cat: union[number]; active: boolean }> = ({
   return (
     <div
       className={`${
-        active ? ' -my-2.5 rounded-3xl bg-white py-2.5 shadow-md' : 'my-5'
+        (active || mouseOver) && cat != 'Events'
+          ? '-my-2.5 mb-2.5 rounded-3xl bg-white py-2.5 shadow-md'
+          : (active || mouseOver) && cat == 'Events'
+          ? '-my-2.5 rounded-3xl bg-white py-2.5 shadow-md'
+          : cat != 'Events'
+          ? 'mb-5'
+          : ''
       }`}
+      onMouseOver={handleMouseOver}
+      onMouseOut={handleMouseOut}
     >
       <div
         className=" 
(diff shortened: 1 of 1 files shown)


src/components/SidebarItems.tsx
```diff
@@ -48,15 +48,33 @@ const SidebarItems: FC<{ cat: union[number]; active: boolean }> = ({
...
     <div
       className={`${
-        active ? ' -my-2.5 rounded-3xl bg-white py-2.5 shadow-md' : 'my-5'
+        (active || mouseOver) && cat != 'Events'
+          ? '-my-2.5 mb-2.5 rounded-3xl bg-white py-2.5 shadow-md'
+          : (active || mouseOver) && cat == 'Events'
+          ? '-my-2.5 rounded-3xl bg-white py-2.5 shadow-md'
+          : cat != 'Events'
+          ? 'mb-5'
+          : ''
```
nl32, line 74: This ternary seems very complicated, maybe try using tailwind modifiers to see if there's a way you can reduce it into just one ternary instead of nested.

src/components/SidebarItems.tsx
```diff
@@ -48,15 +48,33 @@ const SidebarItems: FC<{ cat: union[number]; active: boolean }> = ({
   const router = useRouter();
   const route = routeMap[cat];
 
+  const [mouseOver, setMouseOver] = useState(0);
+  const handleMouseOver = () => {
+    setMouseOver(1);
+    console.log(cat);
```
nl32, line 54: I don't think there needs to be a console log here.

src/components/SidebarItems.tsx
nl32: you could also look into changing the whole thing into tailwind modifiers, but i don't know if that'd necessarily make the code better but might be an interesting learning experience/experiment.

nl32 (changes requested):
I left some comments, the code works and I will approve, but there are some things I commented about that I think would be good to play around with before it's merged in.

RubenOlano:
Try looking into [tailwind states](https://tailwindcss.com/docs/hover-focus-and-other-states). Might simplify code a lot more

src/components/SidebarItems.tsx
```diff
@@ -65,13 +64,11 @@ const SidebarItems: FC<{ cat: union[number]; active: boolean }> = ({
...
-        (active || mouseOver) && cat != 'Events'
-          ? '-my-2.5 mb-2.5 rounded-3xl bg-white py-2.5 shadow-md'
-          : (active || mouseOver) && cat == 'Events'
-          ? '-my-2.5 rounded-3xl bg-white py-2.5 shadow-md'
-          : cat != 'Events'
-          ? 'mb-5'
-          : ''
+        active || mouseOver
+          ? `-my-2.5 rounded-3xl bg-white py-2.5 shadow-md ${
+              cat == 'Events' ? '' : 'mb-2.5'
```
nl32, line 69: tailwind has a [last](https://tailwindcss.com/docs/hover-focus-and-other-states#first-last-odd-and-even) modifier that should make it work regardless of which item is the last one and wouldn't need a ternary. may need to be something like "mb-2.5 last:-mb-2.5"

nl32 (changes requested):

src/components/SidebarItems.tsx
CHarris0812: Thanks for the suggestion. That definitely made it a lot simpler.

src/components/SidebarItems.tsx
CHarris0812: I've messed around with it a bit and I think switching it all to tailwind modifiers makes it more complicated. Not sure if thats just because I don't have a lot of experience with this stuff though.

nl32 (dismissed):
looks good to me 🙂

src/components/SidebarItems.tsx
```diff
@@ -48,28 +48,45 @@ const SidebarItems: FC<{ cat: union[number]; active: boolean }> = ({
   const router = useRouter();
   const route = routeMap[cat];
 
+  const [mouseOver, setMouseOver] = useState(0);
```
RubenOlano, line 51: I would recommend using a boolean here for clarity

RubenOlano (changes requested):
Minor nitpick, other than that it looks good

src/components/SidebarItems.tsx
CHarris0812: Just made this change, should be finished.

RubenOlano (approved):
Looks good 👍🏽

PR #176, opened 2023-11-07
JUP-2 Slide-Out Menu for Navigation

danielsuhcompsci:
I moved the category variables to their own file for use by the Sidebar and MobileNav. I also moved the Icons.tsx file to be under /icons. The mobile nav should be ready, and I could also add an animation if needed.


RubenOlano (approved):
Looks good 👍🏽

KennethImphean (approved):
It looks great! :)

PR #177, opened 2023-11-07
Event search

MistahSanta:
Add a functional search bar for events. Note: I did not add a onClick for when the user click it because we do not have event page for each events. 

My approach: 
- Add an event api call to search and sort event by name that uses ilike to compare the event names. 
- Call this event api to the search bar to display the events and prioritize upcoming event by sorting the query.


src/server/api/routers/event.ts
```diff
@@ -80,4 +85,20 @@ export const eventRouter = createTRPCRouter({
       throw e;
     }
   }),
+  byName: publicProcedure.input(byNameSchema).query( async ({input, ctx}) => {
+    const {name, sortByDate} = input; 
+
+    try { 
+      const events = await ctx.db.query.events.findMany({
+        where: (event) => ilike(event.name, `%${name}%`),
```
nl32, line 93: We should maybe filter out events that have already ended, but I'll leave that decision to the next reviewer

nl32 (changes requested):

RubenOlano:
Honestly, now that we have an actual events page you can add that in. Easiest way to do this would be to wrap the entire thing in a Link Component

src/server/api/routers/event.ts
RubenOlano: I think it should prioritize upcoming events but for now we can keep ended events just in case people want to look at older events

nl32: Yeah, that makes sense, at some point we should maybe set a limit for how far back it would show but that's definitely out of scope of this pr.

PR #178, opened 2023-11-07
JUP-20 new events page

nl32:
New events page with filters and a list and grid view.
The api for filtering is very messy and super coupled with the frontend filterState. any comments regarding cleaning this up or decoupling it would be greatly appreciated.


package.json
```diff
@@ -39,6 +43,7 @@
     "@tailwindcss/typography": "^0.5.9",
     "@types/eslint": "^8.21.3",
     "@types/jest": "^29.5.5",
+    "@types/luxon": "^3.3.3",
```
RubenOlano, line 46: Remove if not being used

src/components/Sidebar.tsx
```diff
@@ -49,7 +49,7 @@ const Sidebar = () => {
           ))}
         </div>
       </div>
-      <div className="w-full py-5 pl-5">
+      <div className="w-full px-5 py-5">
```
RubenOlano, line 52: Is there a reason we're not just using `p-5`?

src/server/api/routers/event.ts
```diff
@@ -13,6 +15,22 @@ const byDateRangeSchema = z.object({
...
+  startTime: z.union([
+    z.literal('now'),
+    z.object({ days: z.number() }),
+    z.custom<DateRange>((val) => isDateRange(val)),
+  ]),
+  order: z.union([
+    z.literal('soon'),
+    z.literal('later'),
+    z.literal('shortest duration'),
+    z.literal('longest duration'),
```
RubenOlano, line 28: Is there a reason why you're using a union of literals over a zod enum?

src/components/events/EventTimeAlert.tsx
```diff
@@ -0,0 +1,54 @@
...
+      {children}
+    </div>
+  );
+};
+const EventTimeAlert = ({ event }: EventTimeAlertProps) => {
+  const now = new Date();
+  const start = event.startTime;
+  const hourDiff = differenceInHours(start, now);
+  if (event.startTime < new Date()) {
+    if (event.endTime < new Date()) {
```
RubenOlano, line 29: Use `now` rather than creating a bunch of dates

RubenOlano (changes requested):
Minor stuff along side using query params for different views & filters

src/components/Sidebar.tsx
nl32: no, just overlooked it 🙂

src/components/events/EventTimeAlert.tsx
```diff
@@ -25,8 +25,8 @@ const EventTimeAlert = ({ event }: EventTimeAlertProps) => {
   const now = new Date();
   const start = event.startTime;
   const hourDiff = differenceInHours(start, now);
-  if (event.startTime < new Date()) {
-    if (event.endTime < new Date()) {
+  if (event.startTime.getTime() < Date.now() ){
+    if (event.endTime.getTime() < Date.now()) {
```
RubenOlano, line 29: I meant use the `now` variable you created

RubenOlano (changes requested):
Other than that its good 👍🏽

src/server/api/routers/event.ts
nl32: changed it to an enum

RubenOlano (changes requested):
Overall amazing 👍🏽

src/app/event/[id]/page.tsx
```diff
@@ -19,38 +20,53 @@ export default async function EventsPage({ params }: Params) {
...
+          rel="canonical"
+          href={'https://jupiter.utdnebula.com/' + event.name}
+          key="canonical"
+        />
+        <meta
+          property="og:url"
+          content={'https://jupiter.utdnebula.com/' + event.name}
+        />
+        <meta name="description" content={`${event.name} - Jupiter`} />
+      </Head>
```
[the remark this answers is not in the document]
Comment (nl32) [reply src/app/event/[id]/page.tsx:36]: didn't see that, all the generate metadata stuff looks right. so ill just remove the Head tag

src/components/events/EventSidebar.tsx
```diff
@@ -0,0 +1,291 @@
...
+                </div>
+                <p className="w-full text-xs font-extrabold text-slate-500">
+                  {value}
+                </p>
+              </div>
+            </RadioGroupItem>
+          ))}
+        </RadioGroup>
+      </div>
+      <div className="flex hidden flex-col space-y-7.5">
```
RubenOlano, line 234: These two conflict, for now remove flex but add a comment saying this is for future implementation

nl32: removed flex and added comment

RubenOlano (dismissed):
🎉 good work!

RubenOlano (dismissed):
Good catch 👍🏽

RubenOlano:
Is the large gap between the org search and the filters necessary?

nl32:
> Is the large gap between the org search and the filters necessary?

no, I just wanted to try and prevent layout shift when you added a club to the search, i removed it though, because you're right, it's too big when there aren't any clubs being filtered.

RubenOlano (approved):

PR #180, opened 2023-11-10
Add tag filtering

RubenOlano:
Adds filtering via tags and uses query params to work with filtering


RubenOlano:
Temporarily, we're just going to use all possible tags but we'll hopefully be able to moderate this in the future

PR #181, opened 2023-11-11
Event search

MistahSanta:
Added onClick functionality that uses useRouter to connect that event id to the dynamic pages of event page


RubenOlano (approved):
LGTM

PR #182, opened 2023-11-14
JUP-26 add event user relation

nl32:
This pr adds:
- a db table to connect users and events.
- a server action to mutate this relation.
  - this doesn't use trpc, hopefully in the future when trpc has more stable support for app router we can use it here.
- a liked events page so that users can see what events they have liked.
- The event cards are now server components.
- events now have a location column


nl32:
one thing is i did add some trpc routes for events, even though i don't think they are used anywhere. I left these here just in case we needed access from the client side.

src/app/events/page.tsx
```diff
@@ -74,7 +74,7 @@ const Events = async ({ searchParams }: { searchParams: searchPams }) => {
   return (
     <main className="pb-10 md:pl-72">
       <EventHeader />
-      <EventView events={events} params={searchParams} />
+      <EventView params={searchParams} events={events} />
```
RubenOlano, line 77: ok bro

src/server/api/routers/event.ts
RubenOlano, line 1: We might need to do some code clean up on this file cause of how large it is

RubenOlano (changes requested):

src/app/events/page.tsx
nl32: I think this just happened because I removed it and added it back lol, cause of how I was doing the composition.

src/app/events/eventView.tsx
```diff
@@ -1,12 +1,11 @@
-'use client';
+'use server';
```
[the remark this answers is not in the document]
nl32: Makes sense, I removed it.

src/server/api/routers/event.ts
```diff
@@ -104,6 +108,28 @@ export const eventRouter = createTRPCRouter({
...
+              id: z.string(),
+              name: z.string(),
+              image: z.string(),
+              description: z.string(),
+            }),
+            liked: z.boolean().optional(),
+            location: z.string(),
+          })
+          .array(),
+      }),
```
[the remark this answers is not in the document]
nl32: I changed the code so that It can infer the output

src/server/api/routers/event.ts
nl32: I feel the size isn't too important, but maybe we open an issue to discuss reorganizing or simplifying some of it?

RubenOlano (approved):
Looks good to merge. The only thing that could be changed would be to use optimistic updates whenever you toggle the like button on events. Up to you if you want to do it in this pr or open a new one

PR #183, opened 2023-11-15
JUP-21 updated sign in page

nl32:
New sign in page matching the current design and is more inline with the nebula styling guide.


src/components/backButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+        <svg
+          xmlns="http://www.w3.org/2000/svg"
+          width="30"
+          height="30"
+          viewBox="0 0 30 30"
+          fill="none"
+        >
+          <path
+            fill-rule="evenodd"
+            clip-rule="evenodd"
```
RubenOlano, line 23: properties should be camelcase

RubenOlano (changes requested):

RubenOlano (approved):

PR #185, opened 2023-11-17
turn tags on clubs into links.

nl32:
changed the tags on club pages into links so that users can find similiar clubs.


RubenOlano (approved):
Good QoL update

PR #187, opened 2023-11-17
JUP-35 Settings page

RubenOlano:
Update the settings page based on the figma design.
Implement procedures that allow for removing orgs and deleting accounts

Not going to merge until we get the join org functionality working so we can test more orgs


RubenOlano:
Test on local since this won't work with credentials

PR #188, opened 2023-11-18
Event page

MistahSanta:
Implemented the Event page using Solomon's design. Ran into some issue with hydration error using the Time Component, so added a check for isLoading using useEffect to prevent that hydration error. CountDown timer is working and accurately count down until start time.


src/app/event/[id]/TimeComponent.tsx
Comment (RubenOlano) [inline src/app/event/[id]/TimeComponent.tsx:1]: For issues with hydration, one thing you can do is pass the date to the function using `date.toISOString()` and then parsing the date in the component using `const date = new Date(dateString)`. That should resolve the server hydration issue you have

RubenOlano (changes requested):
Page looks great!

PR #193, opened 2023-12-13
Directory shows officer information

nl32:
If a club has officers on jupiter then it will list them on the club's directory page.
currently it doesn't show the president on the side or the officer list if there aren't any officers.
you can test this by looking at nebula to see a club with officers and any other club to see what it looks like without.


src/components/OrgInfoSegment.tsx
```diff
@@ -1,8 +1,14 @@
+'use server';
 import Image from 'next/image';
 import React, { type FC } from 'react';
 import type { SelectClub as Club } from '@src/server/db/models';
+import { api } from '@src/trpc/server';
 
-const OrgInfoSegment: FC<{ club: Club }> = ({ club }) => {
+const OrgInfoSegment: FC<{ club: Club }> = async ({ club }) => {
+  const officers = await api.club.getOfficers.query({ id: club.id });
```
RubenOlano, line 8: This will throw an error if not logged in and cause the page to not load

RubenOlano (changes requested):
Looks good so far. Though images for each officer should be left aligned

src/components/OrgInfoSegment.tsx
nl32: I was going to change the api route to public, but I decided against that, so I added a new route to get info specifically for the directory, this has the added benefit that the info segment doesn't need to do any client side fetching any more.

nl32:
I fixed the alignment and changed the way it fetches the officers.

RubenOlano (approved):
Looks good 👍🏽

PR #194, opened 2023-12-18
Update org grid for infinite query :)

RubenOlano:
Initial work for infinite query. Might make it update on scroll if we think that looks better


nl32 (approved):
lgtm, I will probably try and do something similiar for the events page as well.

RubenOlano:
> lgtm, I will probably try and do something similiar for the events page as well.

💀 Good luck

PR #195, opened 2023-12-22
Admin dash

RubenOlano:
Draft of admin dashboard

RN only me and @Jake3231 should be able to access it but you can just add yourself to the admin table if you want to look at it. Click on dropdown on profile once admin. Currently only org stuff is implemented


Jake3231:
@RubenOlano is this awaiting review?

RubenOlano:
> @RubenOlano is this awaiting review?

This was mainly just a look at what we might need in a dashboard component. It's not complete but could use some looking over to make sure that the admin features are implemented correctly and if anything else needs adding. Only other thing I'm thinking about adding is somewhere to view feedback responses

PR #197, opened 2024-01-25
fix - all club members showing up as officers

nl32:
When I updated the way the directory pages fetch org info, when it would fetch officers it would fetch all members instead of filtering it to only officers. This pr fixes this issue.


RubenOlano (approved):
Good catch 👍🏽

PR #198, opened 2024-01-30
Add Images to Clubs

nl32:
Uses a cloudflare r2 bucket to host club images from soc data, shows up in the directory and in the info segment.

Also adds a new column for the profile image, fallbacks to the image column when there isn’t a profile image.


RubenOlano (dismissed):
Looks really good 👍🏽

RubenOlano (approved):

PR #201, opened 2024-02-14
Fix/minor bugs

MistahSanta:
Fixed some bugs that I noticed while looking at the website. 

1) For an event card, truncate the description if there is a long of text that overflow in the card 
image

2) If a org name is long, the org name in the header is large and cover the entire header. Add code to make text smaller if the org name is long 
3) Remove old events components that was used last semester but completely abandon 

image

image


RubenOlano (approved):
Good work 👍🏽

PR #202, opened 2024-02-22
Changes to OrgDirectoryGrid

KennethImphean:
Want to make margins smaller to show 3 orgs per row in directory grid, but resolution is an issue.

### DIFF
--- a/src/components/OrgDirectoryGrid.tsx
+++ b/src/components/OrgDirectoryGrid.tsx
@@ -14,7 +14,7 @@ const OrgDirectoryGrid: FC<Props> = async ({ tag }) => {
   const session = await getServerAuthSession();
 
   return (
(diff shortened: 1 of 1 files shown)


Jake3231:
@KennethImphean I'd suggest updating the title to something more specific to this change, such as "show 3 orgs per row in directory grid" or "Reduce org grid margins"

Also, is it expected to include the README changes in this PR?

KennethImphean:
@Jake3231 I'll change the pull request later to me more specific, this draft was only to show Ruben how the changes would look on his machine because my resolution is bigger on my laptop. I just wanted to make sure that the org grid looked similar on different resolutions. Also, yes the README changes are for newer people to set up ENUMs for the tables in Supabase.

RubenOlano:
Margins are still kinda weird on my end. I think its mostly grid that's causing this. Best thing I can think of is using flex with justify-between or one of those variants so that the cards take up all the available space that they can

Jake3231:
How is this looking?

PR #211, opened 2024-03-01
Jup 44 change heart in events to plus

ishaanthenerd:


.env.example
RubenOlano, line 1: Don't remove this file

RubenOlano (changes requested):
Looks good overall, just the env file thing needs to be fixed

RubenOlano (approved):

PR #216, opened 2024-04-01
JUP-52-make-carousel-bigger

ishaanthenerd:


RubenOlano (dismissed):
Looks good 👍🏽

nl32:
I mean it looks good but I can't tell a difference between this pr and dev. I feel like there was some design specific change that was wanted but I don't remember what exactly.

nl32:
I think it was that Solomon wanted the carousel to fill the whole space horizontally, cause currently if you widen the page enough there is additional white space, @RubenOlano correct me if I'm wrong.

ishaanthenerd:
Yep, that's what Solomon told me.

nl32 (changes requested):
in that case i'm still seeing the issue, there still seems to be white space on the deployment. Ideally it would be the same width as the tag section, which should be as wide as the whole container.

Screenshot 2024-04-01 at 6 27 59 PM

nl32 (approved):
Wide Carousel! lgtm.

RubenOlano (approved):

PR #220, opened 2024-04-05
fix: callbacks for nextAuth signin

nl32:
protected routes will now redirect with a callback so that the user is redirected back to the page they came from improving user experience.


RubenOlano (changes requested):
Minor change then it should be fine

PR #221, opened 2024-04-05
fix: auth redirects not working right on prod

nl32:
switched to using the baseURL function instead of the nextauth_url, was causing issues because it didn't have `https://` and so it treated it as a path instead of a full url and that's what broke the redirect.


nl32:
can't really test on preview but the callbackUrl looks right.

RubenOlano (approved):
One more try

PR #222, opened 2024-04-06
fix: use headers to determine base url for auth redirect

nl32:
unfortunate that it's taken so many attempts, good learning though.
now using the `X-Forwarded-host` and `X-Forwarded-proto` headers to determine what the url the client is using is and follow that. unfortunately really only testable in prod.

### DIFF
--- a/src/app/settings/page.tsx
+++ b/src/app/settings/page.tsx
@@ -31,4 +31,3 @@ const Settings = async () => {
 };
 
 export default Settings;
-3;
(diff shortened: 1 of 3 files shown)


nl32:
real issue is that it's messy determining url from the server, but the ux is nicer if it works. this should definitely work unlike the last times, but I'll do some research into potential issues.

RubenOlano (approved):

PR #225, opened 2024-04-18
JUP-47: improve events page

nl32:
Rebuild of the events page that removes filters. There was a long standing issue of figuring out what kind of criteria users might want to filter events by. We struggled to find an answer beyond, events for clubs users have joined, and events users have shown interest in. Filtered event views are available through club listings and the my community page so this added complexity can be removed from the events page.

To achieve this the pr makes these changes:
- New SearchParams utility to make interacting with searchParams in a type-safe way easier
- Removing the event sidebar
- Removing the swappable list and grid views (Always list now)
- Added improved date picker
- styling improvements focused on mobile


KennethImphean (approved):
Good change to not overwhelm the user when looking at events!

RubenOlano (approved):
LGTM

PR #226, opened 2024-04-18
Fix Deployments and update dependencies

nl32:
I don't know why deployments started breaking but they did and by upgrading everything It deploys right. Deployments were breaking even for functionally identical branches to `develop`. As such it was either a vercel issue or some upstream dependency.
Got to love the bleeding edge.

The deployment issue was caused by a trpc call from within Server Component causing a content length mismatch error, which we'd encountered early on in the app router migration for various reasons. 
The source of all these headaches was  having to use the `TRPCProxyClient` in order to make TRPC calls from Server Components. This was always a bit of a hack, normally tRPC exposes a `caller` to be used to invoke TRPC functions serverside, however `caller` couldn't use a callback for it's context, meaning we couldn't use it for Server components. This was changed in TRPC v11, allowing us to use `caller` in server components. This unfortunately needed multiple dependencies bumped up. This should overall be more stable.

## Main Changes
- Upgrade NextJS to v14: no breaking changes, server actions are "stable", was doing it anyways for some server action stuff on a feature branch, though I'd do it here to make sure the changes would work with next 14
- Upgrade TRPC to v11: needed to change Server API
- Upgrade React-Query to v5: TRPC v5 upgrades to v5 so we had to as well

## Additional Changes
- This also fixed the `DYNAMIC_SERVER_USAGE` errors at build time
- serverside TRPC API calls no longer need `.query`
- React-Query v5 breaking changes, main ones that affected the codebase were [onX callbacks being removed from useQuery](https://tkdodo.eu/blog/breaking-react-querys-api-on-purpose), and `isLoading` being changed to `isPending` on mutations, you can see the full list [here](https://tanstack.com/query/v5/docs/framework/react/guides/migrating-to-v5).


next.config.mjs
```diff
@@ -13,9 +13,6 @@ const config = {
       'jupiter.nlmc.workers.dev',
     ],
```
RubenOlano, line 14: Make sure to update this `remotePatterns` for next 14

RubenOlano (changes requested):
Looks good so far, make sure to update any remaining apis that change between next 13 and 14. Image domains as well as getting rid of the `theme` option in the metadata objects

RubenOlano (approved):

PR #230, opened 2024-07-22
JUP-57 Sidebar

nl32:
Refactor the full sidebar and mobile sidebar to use the same component for the actual navigation content.

also moves the admin menu to the sidebar, along with the club manage menu, if a user should have access to these pages.


RubenOlano (approved):
Looks good, only minor nitpick would be that the sidebar does not show up on mobile but I'm pretty sure that's my fault and not yours

PR #231, opened 2024-08-20
Add Vercel analytics

TyHil:
Add Vercel analytics like in trends and planner, also see pr in website-v2: https://github.com/UTDNebula/website-v2/pull/59


nl32 (approved):
Trust, lgtm. changed it to merge into main so we can get analytics on production sooner.

PR #237, opened 2024-09-03
JUP-84 tag improvements

nl32:
this pr introduces a new database view that lets us see all the unique tags and how many clubs there are for each tag, it's default ordering is also by count. This new ordering also applies to the tag listing on the homepage. 

we may want to switch it to a materialized view with triggers as this information shouldn't be updated often, as view can't be indexed in postgres.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,10 @@ $$;
 Currently, drizzle does not automatically create enums for you. You will have to create them manually. This [link](https://orm.drizzle.team/docs/column-types/pg#enum) should give you a good idea of how to create enums in postgres.
 
 </details> </br>
+<details>
+<summary>Tag View</summary>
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -104,7 +104,11 @@ $$;
 
 Currently, drizzle does not automatically create enums for you. You will have to create them manually. This [link](https://orm.drizzle.team/docs/column-types/pg#enum) should give you a good idea of how to create enums in postgres.
 
-</details> </br>
+</details> 
+<details>
+<summary>Tag View</summary>
+  there is a sql query in `src/server/db/tagView.sql` that you need to run in order for tags to work properly.
+</details></br>
```
RubenOlano, line 111: Explain why we need this

RubenOlano (changes requested):
Looks good, change README :)

PR #238, opened 2024-09-08
Jup 78 - Create event edits

BK2004:
- event start and end date only
- event end set to start when start is greater or end is undefined
- event picture select grayed out
- redirects to new event on success


nl32 (changes requested):
real small thing, otherwise lgtm.

RubenOlano (approved):
Good work!

nl32 (approved):
lgtm

PR #239, opened 2024-09-30
JUP-63-Create-Joined-clubs-carousel

ishaanthenerd:
I'm not fully sure how to make the layout of this more appealing (e.g. underneath the left arrow, club cards do not align) but I have the basic carousel in place.


src/components/InfiniteScrollGrid.tsx
ishaanthenerd, line 1: What is the type of cursor on line 18?

PR #240, opened 2024-09-30
Component Reorganization

nl32:
This pr reorganizes the contents of the Components folder so that they are better organized and grouped by the domain they relate to (Header, Directory, etc.)

Some components have been moved into new folders within the components folder, while others have been moved into the app folder. this determination of where things ended up is dependent on where and how the components are consumed.

also `ContactIcons` was moved to the icons folder.

The searchBar has been extensively broken down in order to try and seperate the debounce logic from the styling itself, this should make it easier to reuse the component in places where the debounce & autocomplete behavior isn't wanted. Other changes to the searchBar will come in a later pr.


RubenOlano (changes requested):
Looks good overall, since you're doing this, we should make the switch over to using Club rather than Org now. Other than that, minor import nitpicks

PR #241, opened 2024-10-01
Search bar Submit functionality.

nl32:
expands the search bar api to add a submit button that can either be passed a handler function, through `submitLogic` or behave like a submit button for a form, this function will also be invoked if the user presses enter in the search bar. also tweaks the current search bars to take advantage of this feature.

### DIFF
--- a/src/components/searchBar/index.tsx
+++ b/src/components/searchBar/index.tsx
@@ -1,9 +1,14 @@
-import { SearchIcon } from '@src/icons/Icons';
-import { ComponentProps } from 'react';
+import { RightArrowIcon, SearchIcon } from '@src/icons/Icons';
+import { type ComponentProps } from 'react';
 
-type SearchBarProps = Omit<ComponentProps<'input'>, 'type'>;
+type SearchBarProps = Omit<ComponentProps<'input'>, 'type'> & {
+  submitButton?: boolean;
+  submitLogic?: () => void;
+};
 
 export const SearchBar = (props: SearchBarProps) => {
+  const submitButton = props.submitButton;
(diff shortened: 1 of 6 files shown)


src/components/searchBar/index.tsx
```diff
@@ -12,8 +17,32 @@ export const SearchBar = (props: SearchBarProps) => {
       <input
         {...props}
         type="text"
-        className={`h-10 w-full rounded-full border pl-10 pr-3 focus:outline-none ${props.className}`}
+        className={`h-10 w-full rounded-full border pl-10 ${submitButton ? 'pr-[38px]' : 'pr-3'} focus:outline-none ${props.className}`}
+        onSubmit={() => {
+          console.log('howdy from submit');
```
TyHil, line 22: maybe remove this

src/components/searchBar/index.tsx
```diff
@@ -12,8 +17,32 @@ export const SearchBar = (props: SearchBarProps) => {
...
+        }}
+        onKeyDown={(e) => {
+          if (e.key === 'Enter' && submitLogic) {
```
TyHil, line 25: should this be `typeof submitLogic !== 'undefined'`

TyHil (approved):

src/components/searchBar/index.tsx
nl32: done.

src/components/searchBar/index.tsx
nl32: does the same thing but i agree.

TyHil (approved):

PR #242, opened 2024-10-01
Updated the about page

Ninad-S:


KennethImphean (changes requested):
When you import a component from another folder, make sure to include the entire directory path! The vercel deployment is failing because it's looking for Header in the components folder, but it's in the header folder. Also make sure to do npm run dev and npm run build to make sure the deployment is good on your machine before committing and making your pull request.

PR #244, opened 2024-10-01
Jup 61 club search page

CHarris0812:
Made it so that after pressing enter in the search bar it shows all the clubs that match that search.


src/server/api/routers/club.ts
```diff
@@ -64,6 +64,17 @@ export const clubRouter = createTRPCRouter({
 
     return clubs.slice(0, 5);
   }),
+  byNameNoLimit: publicProcedure
```
nl32, line 67: LMAO, we should probably just rewrite the byName api to accept a limit. don't worry about it on this pr though

nl32 (approved):
There's some changes that need to be made to how the searchBar is, the overengineering seems to have really broken it. We'll leave that to a different pr though.

PR #248, opened 2024-10-08
Add Google Analytics

TyHil:
Added google analytics based off of this: https://medium.com/readytowork-org/google-analytics-in-next-js-a26cc2b28db5

Works for Trends, but only in prod. Which is good I suppose


RubenOlano (approved):
LGTM 👍🏽

PR #250, opened 2024-10-18
Made the Contacts into its own section

SGcoder120:
Now the Contacts bar is its own section.


nl32 (changes requested):
design should match the figma.
Screenshot 2024-10-22 at 2 56 24 PM

nl32:
closed: duplicate pr

PR #252, opened 2024-10-23
JUP-71 Backend Changes

BK2004:
- Added officers table for listed officers
- Removed title field from userMetadataToClubs and all references to it


src/server/db/schema/officers.ts
```diff
@@ -0,0 +1,24 @@
...
+    clubId: text('club_id')
+      .notNull()
+      .references(() => club.id),
+    name: text('name').notNull(),
+    position: text('position').notNull(),
+    image: text('image').default('/nebula-logo.png').notNull(),
+    isPresident: boolean('is_president').default(false).notNull(),
+  },
+  (table) => ({
+    pk: primaryKey(table.clubId, table.id),
```
RubenOlano, line 18: Why are we doing composite key here?

src/server/db/schema/officers.ts
```diff
@@ -0,0 +1,24 @@
+import { relations } from 'drizzle-orm';
+import { pgTable, text, boolean, primaryKey } from 'drizzle-orm/pg-core';
+import { club } from './club';
+
+export const officers = pgTable(
+  'officers',
+  {
+    id: text('id').notNull(),
```
RubenOlano, line 8: Need to set some default value.
Maybe `uuid`?

RubenOlano (commented):
Couple of questions here let me know if I'm just being dumb

src/server/db/schema/officers.ts
BK2004: Now that I'm looking at it, there really isn't a reason for a composite key. I'll get rid of that.

RubenOlano: Sounds good. I was thinking that maybe it was supposed to be linked with the `userId` but it's not a column in the table so prob just leave it out

src/server/db/schema/officers.ts
```diff
@@ -0,0 +1,18 @@
+import { relations } from 'drizzle-orm';
+import { pgTable, text, boolean, uuid } from 'drizzle-orm/pg-core';
+import { club } from './club';
+
+export const officers = pgTable('officers', {
+  id: uuid('id').defaultRandom().notNull().primaryKey(),
```
RubenOlano, line 6: Unnecessary `notNull`

RubenOlano (changes requested):
One final nitpick

RubenOlano (approved):
LGTM

nl32 (approved):
LGTM

PR #261, opened 2024-11-14
Add format script

TyHil:
Added a formatting script instead of `npx prettier --write .`
Updated the GitHub action to run that script as well as the lint script.

Also, fully removed husky and commitlint as they weren't being used. If this wasn't needed I can revert

Need to update branch protection rules to reflect the new GitHub action once merged: https://github.com/UTDNebula/jupiter/settings/branch_protection_rules


nl32 (approved):
LGTM

PR #265, opened 2025-02-12
Jup 125 removed leadership img

jordanjoelson:
Removed image and adjusted width of description

### DIFF
--- a/src/components/club/listing/ClubInfoSegment.tsx
+++ b/src/components/club/listing/ClubInfoSegment.tsx
@@ -47,7 +47,7 @@ const ClubInfoSegment: FC<{
             </p>
           </div>
         </div>
-        <div className="w-full">
+        <div className="w-full md:w-2/3">
           <p className="whitespace-pre-wrap text-slate-700">
             {club.description}
           </p>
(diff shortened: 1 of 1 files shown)


src/components/club/listing/ClubInfoSegment.tsx
```diff
@@ -46,13 +46,6 @@ const ClubInfoSegment: FC<{
               <div className="flex flex-col justify-center align-middle">
                 {club.officers.map((officer) => (
                   <div className="mt-5 flex flex-row" key={officer.id}>
-                    <Image
```
nl32, line 49: 👍

src/components/club/listing/ClubInfoSegment.tsx
```diff
@@ -34,7 +34,7 @@ const ClubInfoSegment: FC<{
             </p>
           </div>
         </div>
-        <div className="w-full">
+        <div className="w-full md:w-2/3">
```
nl32, line 37: what problem led you to this solution? I'm curious as to if there might be a better solution.

nl32 (changes requested):
Main problem is resolved, want to have further discussion over the spacing of whole component.

BK2004 (approved):

PR #270, opened 2025-02-25
Event Date Picker Improvements

CHarris0812:
Small changes. Made it so that the native date picker is used on mobile. I also removed the text entry ability in the field because it was all funky and seemed unnecessary.

### DIFF
--- a/src/app/events/eventView.tsx
+++ b/src/app/events/eventView.tsx
@@ -2,7 +2,8 @@
 import DateBrowser from '@src/components/events/DateBrowser';
 import { type eventParamsSchema } from '@src/utils/eventFilter';
 import useSyncedSearchParams from '@src/utils/useSyncedSearchParams';
-import { type ReactNode } from 'react';
+import { type ReactNode, useEffect, useState } from 'react';
+
 type Props = {
   children: ReactNode;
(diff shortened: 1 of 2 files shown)


src/app/events/eventView.tsx
```diff
@@ -2,22 +2,46 @@
...
 type Props = {
   children: ReactNode;
   searchParams: eventParamsSchema;
 };
 
 const EventView = ({ children, searchParams }: Props) => {
   const [params, setParams] = useSyncedSearchParams(searchParams, '/events');
+  const [isMobile, setIsMobile] = useState(false);
+
+  useEffect(() => {
```
nl32, line 16: Let's see if theres a way to make this more concise.

src/app/events/eventView.tsx
```diff
@@ -2,22 +2,46 @@
...
+    const checkMobile = () =>
+      setIsMobile(window.matchMedia('(max-width: 768px)').matches);
+    checkMobile();
+    window.addEventListener('resize', checkMobile);
+    return () => window.removeEventListener('resize', checkMobile);
+  }, []);
```
nl32, line 22: check dependency array.

nl32 (changes requested):

PR #275, opened 2025-04-02
JUP-122: Add make president feature to editListedOfficersForm

BK2004:
- Each officer under manage officers has the option to "make president"
- Only one officer at a time can be president


nl32 (approved):
LGTM!

PR #277, opened 2025-04-18
Google analytics only on prod

TyHil:
Only initialize google analytics on prod. Just made this change to the other Nebula sites.

Also, changed prettier to not care about line endings. Git can handle that.


nl32 (approved):
lgtm

PR #280, opened 2025-04-29
Tailwind upgrade to v4

nl32:
This migrates to tailwind v4 which has new features and simpler config. everything has been migrated, some utilities have been renamed for the new upgrade. check [https://tailwindcss.com/docs/upgrade-guide](here) for breaking changes


TyHil:
Just looked over this since I was doing similar work recently. In this PR and on dev/main I think Inter is not being used but other than that all looks good

PR #284, opened 2025-05-13
Jup 141 typesafe helpers for cloud storage api

TyHil:
Based off of #283 

Here are some basic wrappers on the the APIs cloud storage endpoints. I'm new to this tRPC stuff so lmk any changes to make.

I think I could have zod do runtime response type checking if we want. And I'm not sure how we want to do error handling, as in if it's ok for an error to just be thrown in the router.


TyHil:
I have realized that I don't need the kind of error handling I'm doing here. Will rework

PR #286, opened 2025-05-25
Add Sentry and Bug Report Button

TyHil:


nl32:
looks good, however it seems to add a whole minute to the build times, which seems less than ideal, this may be a hit we can't avoid but thought you should be aware.

TyHil:
Just disabled it on not production deployments. Seems faster

PR #287, opened 2025-07-14
Club match

TyHil:
"Find your club match! Generate club recommendations based on a simple form."

image
image

Based on https://github.com/will-horiszny/InstantMatch


TyHil:
@BK2004 and @nl32 up to y'all what you want to do with this. I think its in a good state functionality-wise but design-wise it could be a good thing to look at and rework at the start of the semester. I'm still waiting for some pricing data to come into GCP to see how expensive it will be

PR #294, opened 2025-08-15
Search Overhaul

nl32:
Adds:
- tag searching (~~currently ORs tags~~ now uses AND)
- text search on clubs
- suggests top tags for users to filter by

Styling still needs some refinement but I believe functionality is solid,
Some tweaking to the directory transition behaviors and debounce timing are probably needed.
Opening this as a draft to collect feedback.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,6 +24,7 @@
         "@sentry/nextjs": "^9.22.0",
         "@tanstack/match-sorter-utils": "^8.8.4",
         "@tanstack/react-query": "^5.74.3",
+        "@tanstack/react-query-devtools": "^5.91.0",
         "@tanstack/react-table": "^8.11.0",
         "@tanstack/react-virtual": "^3.0.1",
         "@trpc/client": "^11.1.0",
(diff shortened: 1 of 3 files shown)


src/server/api/routers/club.ts
```diff
@@ -337,4 +361,64 @@ export const clubRouter = createTRPCRouter({
...
+                  should =>ARRAY[
+                    paradedb.boost(2.0,paradedb.match('name',${input.search},distance=>1)),
+                    paradedb.boost(1,paradedb.match('description',${input.search},distance=>1)),
+                    paradedb.boost(1.5,paradedb.match('tags',${input.search},distance=>1))
+                  ])`
+              : undefined,
+          ),
+        )
+        .orderBy(input.search ? sql`paradedb.score(id) DESC` : club.name);
+      console.log(query.toSQL());
```
BK2004: remove this

BK2004 (changes requested):
Looks good. Note, we may want to remove the # from the pills for tags.

src/server/api/routers/club.ts
nl32: done

TyHil (approved):

PR #324, opened 2025-09-25
Search Bar Anchor to Top of Screen

LandRtx:
Home page search bar becomes sticky to the top of the screen when the user scrolls past it.

image

Possible improvements:
- Make search bar hang below the top rather than right on the edge. Needs a smooth transition or a way to hide the drop as you scroll past.
- Make search bar function at any point on the page, rather than jumping to the top when a query is typed


BK2004 (approved):
This works well. One thing to note is that searching jumps the user straight back to the top, which is not ideal, so we might want to look into changing searching to be only on pressing enter or something of the sort.

PR #346, opened 2025-10-24
Add Material UI

TyHil:
Adds theme and imports for Material UI and converts one components, the profile photo and drop down, to use MUI.


TyHil:
@nl32 @BK2004 you can't tell me this doesn't look better

image

image

PR #367, opened 2025-10-31
Upload images tab for clubs (see if this is what you were thinking)

seean7:
This is what I was thinking for the upload images tab let me know if this works for y'all and this is what you wanted


src/app/manage/[clubId]/Uploads/page.tsx
Comment (BK2004) [inline src/app/manage/[clubId]/Uploads/page.tsx]: Make sure to remove this page before making final PR

BK2004 (changes requested):
This looks good!

Issue #314, opened 2025-09-09
Fix homepage loading state when searching


BK2004:
Currently, the homepage shifts up and down when searching due to loss of data when searching.

Look at this and try to solve this problem: [Placeholder Query Data](https://tanstack.com/query/latest/docs/framework/react/guides/placeholder-query-data#what-is-placeholder-data)

Use a skeleton to render a few empty clubs cards: https://mui.com/material-ui/react-skeleton/

PR #368, opened 2025-11-02
Fix homepage loading state when searching (#314)

SaivishaalSureshkannan:
This fixes issue #314 where the user keeps getting redirected to the top of the page when they enter a new query(sequence of letters or letter that hasn't been searched before in the session). This update ensures the user remains at their current scroll position unless the query returns no results. Also, Skeleton loaders have been added to both the search results and club cards to maintain consistent layout height during data fetching. 

Changes made:
- Added placeholderData to HomePageSearchBar and ClubDirectoryGrid to keep previous data during fetch
- Added skeleton loaders
- Scroll-to-top behavior now only triggers if the query entered by the user has no results.


BK2004 (approved):
Looks good!

nl32 (approved):
lgtm

PR #369, opened 2025-11-03
feat(ui): make ClubCard fully clickable and remove Learn More button

sy-pk:
### Context
Only the Learn More button was clickable. 
Making the entire card clickable provides a better UX.

### Changes
- Made the entire `ClubCard` clickable to navigate to the detail page
- Removed “Learn More” button

### Preview
| Before | After |
|--------|--------|
|Before|After|


sy-pk:
I just checked this again.
Clicking the Join button doesn’t navigate anywhere.
The link preview still appears on hover, but the click itself doesn’t trigger navigation.
Everything seems to be working as expected!

BK2004:
It seems that it doesn't navigate if the join button is disabled but it does when active, which is why I missed the issue originally.

sy-pk:
I wasn’t logged in earlier, so I missed checking that case. I’ve fixed it and opened PR #373!

PR #372, opened 2025-11-03
feat(clubcard): add heart button with white icon

sy-pk:
### Context
Added a heart (like) button to the ClubCard so users can save clubs they’re interested in and view them together later.

### Changes
- Added heart icon (`Favorite`, `FavoriteBorder`) to the top-right corner of `ClubCard`
- Implemented toggle state with subtle hover and active scaling
- Matched existing shadow style for visual consistency with other white icons

### Preview

Before
Before

After
After

### DIFF
--- a/src/components/club/ClubCard.tsx
+++ b/src/components/club/ClubCard.tsx
@@ -1,13 +1,22 @@
+import { Favorite, FavoriteBorder } from '@mui/icons-material';
 import { type Session } from 'next-auth';
 import Image from 'next/image';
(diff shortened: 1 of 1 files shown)


sy-pk:
The Join button may not be very relevant during the exploration stage, so it could be better to replace it with other information such as member count or founding year.
It would be good to discuss this together in tomorrow’s meeting. If we agree to remove it, we should also align on what to show and request updated visuals from the design team.

PR #373, opened 2025-11-03
fix: prevent ClubCard navigation when clicking JoinButton

sy-pk:
### Context 
The Join button inside the ClubCard was still triggering navigation to the club detail page when clicked.

### Changes
- Added `stopPropagation()` and `preventDefault()` inside the JoinButton click handler to prevent navigation.

### DIFF
--- a/src/components/club/JoinButton.tsx
+++ b/src/components/club/JoinButton.tsx
@@ -10,7 +10,6 @@ type JoinButtonProps = {
   isHeader?: boolean;
   isJoined?: boolean;
   clubID: string;
-  onClick?: (e: React.MouseEvent) => void;
 };
 
 const JoinButton = ({
@@ -18,7 +17,6 @@ const JoinButton = ({
   session,
   isJoined,
   clubID,
-  onClick,
 }: JoinButtonProps) => {
   const api = useTRPC();
   const mutation = useMutation(api.club.joinLeave.mutationOptions());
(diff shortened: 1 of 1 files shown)


src/components/club/ClubCard.tsx
```diff
@@ -45,7 +45,14 @@ const ClubCard: FC<Props> = ({ club, session, priority }) => {
 
         <div className="m-5 mt-auto flex flex-row space-x-2">
           <div onClick={(e) => e.stopPropagation()}>
```
TyHil: Definitely a nitpick but is stopPropagation still necessary here

src/components/club/ClubCard.tsx
```diff
@@ -45,7 +45,14 @@ const ClubCard: FC<Props> = ({ club, session, priority }) => {
...
-            <JoinButton session={session} clubID={club.id} />
+            <JoinButton
+              session={session}
+              clubID={club.id}
+              onClick={(e) => {
```
TyHil: and do we need to pass in `onClick` if join button already does `e.stopPropagation()` and `e.preventDefault()`? We could pick either having to pass it in or always having it done in the component.

TyHil (commented):
Couple small things

src/components/club/ClubCard.tsx
sy-pk: I updated it so that the event handling happens inside the JoinButton component instead. Does that look okay?

PR #392, opened 2025-11-25
Update Club Page

### LINKED ISSUE #354: Update Club Page
# Overview

Users won't know what to make of an 'Active' section. Also, the name and description headers are redundant since the name is displayed at the top of the club page, and description on the right.

# Proposal
- Change 'Active' to 'Last Updated', and use a date instead of True/False
- Remove name section on the left
- Remove description header

Should look like this:

Image

nl32:
closes #354  
- Moves club tags to description
- removes club name from description
- shows update field if club has ever been updated on the platform (check nebula for example)
- only shows contact information segment if clubs have any
- only shows events for clubs that aren't sourced from soc

## looking for feedback
Added a new section at the bottom of the page for any club marked as coming from SOC providing our discord and instagram as contact points to help get club information updated. This should make it easier for organizations that look themselves up on clubs to get setup on the platform.

### DIFF
--- a/src/components/club/listing/ClubNotClaimed.tsx
+++ b/src/components/club/listing/ClubNotClaimed.tsx
@@ -11,14 +11,14 @@ export const ClubNotClaimed = () => {
           reach out on our{' '}
           <a
             href="https://discord.utdnebula.com/"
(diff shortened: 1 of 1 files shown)


TyHil (approved):
🎉

PR #394, opened 2025-11-25
Full size homepage with scroll on interaction

TyHil:
Goes back to the full page homescreen but scrolls on any interaction with the search bar or tags, only the first time.

Waiting on discussion in #clubs-general


TyHil:
Discussion in #clubs-general seems to think this is good, should be ready @BK2004 !

PR #398, opened 2025-11-26
database migration and branching improvements

nl32:
This as well as setting up the neon vercel integration will let us generate new database branches for the different deployments, meaning no more deployment problems because changes haven't been applied to the upstream databases.

This pr itself contains new config for database migrations and docs for generating/applying them.
Once the pr has been reviewed I will setup the integrations so that database changes on branches can get reflected in the deployments. It's a good step in making testing and database management less manual.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,5 +64,7 @@ When working on a new feature, please create a new branch with the following nam
 git checkout -b feature/<feature-name>
 ```
 
+If your changes involve any changes to the database schema please read `/docs/db_migrations.md`
+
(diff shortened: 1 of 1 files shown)


nl32:
> So stepping through the workflow for a member, let me know if I'm totally off:
> 
> * This wouldn't do anything to change the DB they're connected to locally
> * Once their branch is deployed to Vercel db changes they made would be on a Neon branch
> * Once their branch is merged into develop, would the migration for develop be done automatically?

1. It would, it replaces `npm run drizzle:push` with `npm run drizzle:generate` and `npm run drizzle:migrate` this is so the other 2 steps can work.
2. yes
3. and yes

Issue #353, opened 2025-10-27
Google Calendar


TyHil:
# Overview

Allow orgs to add their Google Calendars to populate events in Clubs.

https://developers.google.com/workspace/calendar/api/v3/reference/events/list

PR #403, opened 2025-12-02
Migrate from next-auth to better auth

nl32:
Makes the switch from Next-Auth.js to Better Auth, This is a decision I was hesitant to make, however due to issues I've encountered when working on #353 I've decided it would be prudent to make this migration before we have live users.

The reasoning behind the switch:
- better support, next-auth is pretty much dead except for security updates
- better documentation, if you take a look at the [next-auth documentation](https://next-auth.js.org/getting-started/introduction) and the [Better Auth documentation](https://www.better-auth.com/docs/introduction) you can see a clear difference in quality and detail.
- better [extensibility](https://www.better-auth.com/docs/concepts/plugins)
- preview deployment authentication support (this will be a seperate pr)
- was encountering an issue with next-auth not updating access/refresh tokens from google sign in.

---
Besides changing environment variable names all other next-auth to better auth changes should be accounted for by this pr.


TyHil (dismissed):
Works locally for me and looks good.

PR #404, opened 2025-12-02
migrate to the eslint cli from next lint

nl32:
migrate to the eslint cli since next lint is being deprecated with next 16.
also means the linting script doesn't need environment variables for the app.


TyHil (approved):
LGTM

PR #469, opened 2025-12-17
431 refactor club match form

AbhiramTadepalli:
removed Controller, using tanstack's useForm, using form.Field wrappers. Use subscribe when i need conditional rendering


src/app/club-match/ClubMatch.tsx
```diff
@@ -208,82 +209,93 @@ const SelectMultipleInput = ({
...
 
   const api = useTRPC();
   const router = useRouter();
+
   const editData = useMutation(
     api.ai.clubMatch.mutationOptions({
       onSuccess: () => {
         router.push('/club-match/results');
       },
+      onError: (err) => {
```
TyHil: I think we can also remove this since the tanstack validator will catch things that used to make this run

src/app/club-match/ClubMatch.tsx
```diff
@@ -296,202 +308,281 @@ const ClubMatch = () => {
...
+                  ]}
+                  errors={errors}
+                  field={field}
+                />
+              )}
+            </form.Field>
           </div>
         </div>
+
         <Button
```
TyHil: We can use the `FormButtons` like in other forms:
```
<div className="flex flex-wrap justify-end items-center gap-2">
  <form.AppForm>
    <form.FormResetButton />
  </form.AppForm>
  <form.AppForm>
    <form.FormSubmitButton />
  </form.AppForm>
</div>
```
Probably with an optional prop added to `FormSubmitButton` to change the text from "Save" to "Find Clubs"

src/app/club-match/ClubMatch.tsx
```diff
@@ -296,202 +308,281 @@ const ClubMatch = () => {
       <form
```
TyHil: If we want to use `Panel` inside this like with other forms that would be good, for standardization. We could keep the same `max-2-3xl` tho if it looks good.

src/app/club-match/ClubMatch.tsx
```diff
@@ -59,8 +59,7 @@ const TextInput = ({ id, label, disabled, field }: SharedInputProps) => {
         error={shouldShowError}
         helperText={
           shouldShowError
-            ? field.state.meta.errors.map((err) => err?.message).join('. ') +
-              '.'
+            ? field.state.meta.errors.map((err) => err?.message).join('. ')
```
TyHil: we have periods after errors on all the other forms

TyHil (changes requested):
this is going to have some effects on other forms that need to be addressed

TyHil (approved):
Looking good!

PR #472, opened 2025-12-17
448 store club match question answers

AbhiramTadepalli:
Is it fine to use the `ClubMatchResponses` type in `ClubMatch.tsx` as a prop for the db response?

### DIFF
--- a/src/app/club-match/results/RedoClubMatchButton.tsx
+++ b/src/app/club-match/results/RedoClubMatchButton.tsx
@@ -2,8 +2,8 @@
 
 import { RestartAlt } from '@mui/icons-material';
 import { Button } from '@mui/material';
-import { useState } from 'react';
 import Link from 'next/link';
+import { useState } from 'react';
 
 const RedoClubMatchButton = () => {
   const [loading, setLoading] = useState(false);


src/app/club-match/results/RedoClubMatchButton.tsx
```diff
@@ -1,17 +1,29 @@
...
+
   return (
     <Button
       variant="contained"
-      component={Link}
-      href="/club-match"
+      onClick={handleClick}
+      disabled={loading}
       className="rounded-full w-fit normal-case self-center"
+      startIcon={
```
TyHil: Use the loading prop instead of disabled:
```
loading
loadingPosition="start"
```

src/app/club-match/results/RedoClubMatchButton.tsx
```diff
@@ -1,17 +1,29 @@
...
+import { useRouter } from 'next/navigation';
+import { useState } from 'react';
 
 const RedoClubMatchButton = () => {
+  const router = useRouter();
+  const [loading, setLoading] = useState(false);
+
+  const handleClick = () => {
+    setLoading(true);
+    router.push('/club-match');
```
TyHil: Does this need to use `router.push`? Would be better to keep it as a link. You can still probably use the onClick to set the loading state.

TyHil (changes requested):

PR #477, opened 2025-12-17
451 user metadata from club match

AbhiramTadepalli:
Just added a db update in `ai.ts`. Also call `userMetadata` on page load for `/club-match` to get the user's major from profile


AbhiramTadepalli:
> can y'all actually test it and make sure it's working correct

@TyHil

PR #480, opened 2025-12-17
452 remove clubs user is in

AbhiramTadepalli:
filtered out the clubs a user is in before feeding it to gemini


TyHil:
Waiting on #469

Issue #379, opened 2025-11-07
Club Card Image Aspect Ratio


sy-pk:
### What did you expect to happen?

The club logo should keep its original square ratio on all screen sizes. It should resize normally without looking stretched or squished on mobile.

### What happened?

On smaller screens, the logo image gets stretched horizontally. Looks like the image container has a fixed height, so the aspect ratio breaks when the width changes.

Desktop

<img width="1361" height="656" alt="Image" src="https://github.com/user-attachments/assets/87f2a2e4-0ffd-41ce-b985-3ddccc03bd74" />

Mobile

<img width="365" height="557" alt="Image" src="https://github.com/user-attachments/assets/e92742db-c629-4527-8079-8f64e2e2c56e" />

### Version

Pre-release (Development)

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #493, opened 2025-12-22
Club card image aspect ratio (#379)

sy-pk:
### Changes
- Ensured club card images maintain a square aspect ratio across all screen sizes
- Adjusted logo fitting to avoid cropping and visually align with the card background

### DIFF
--- a/src/components/events/EventCard.tsx
+++ b/src/components/events/EventCard.tsx
@@ -34,7 +34,7 @@ const EventCard = ({ event, view = 'normal' }: EventCardProps) => {
               fill
               src={src}
               alt="event image"
-              className="object-cover object-left"
+              className="object-cover object-center"
             />
           )}
           <div className="absolute inset-0 p-2">


TyHil (changes requested):
Just reviewing for formality. I think we need the same changes on `EventCard` before merging.

sy-pk:
> Just reviewing for formality. I think we need the same changes on `EventCard` before merging.

My concern with using `object-contain` with a white background on `EventCard` is that many events will likely reuse club logos instead of having custom thumbnails (like some of our current events). With a wider aspect ratio, this can create a lot of empty space and look awkward. 
That was part of why I originally wanted to standardize image ratios, but since we decided to move forward with a 2:1 ratio and I believe Issac is already working on it, I’m not sure what the best approach is here. 
I understand the goal of standardizing image behavior, but if we go that route, it might make more sense to also standardize the aspect ratio (e.g. 1:1). Otherwise, keeping `object-cover` for `EventCard` could be visually cleaner than matching the `ClubCard` behavior.

sy-pk:
We could do it that way, but I’m a bit concerned it could cause visual inconsistency. If we treat logo fallbacks and custom thumbnails differently (logo = `cover`, custom image = `contain`), users may notice that some EventCards fill the image area while others have white space, which could be confusing.

Even if we encourage a 2:1 event thumbnail, people likely won’t always export images at exactly that ratio. Using `object-contain` in those cases would introduce noticeable empty space (left/right or top/bottom), which can look awkward.

If a club creates a wide thumbnail for a particular event, `object-cover` should work well when it is close to 2:1, and if it is not exact, `object-cover` is more forgiving and tends to look cleaner overall.

I also tried applying a blurred version of the logo to fill the extra space before, since most logos are close to 1:1 but not all of them are, but it didn’t end up looking as nice as I expected.

Issue #494, opened 2025-12-22
Inconsistent typography scaling across breakpoints in club card


sy-pk:
### What did you expect to happen?

- The club card title and body text maintain a consistent font size across screen sizes
- Responsiveness is handled through layout and spacing, not font-size scaling

### What happened?

- Font sizes change across breakpoints, causing the title and body text to become smaller on larger screens
- This results in reduced readability and inconsistent typography in the club card layout

| Mobile | Desktop |
|--------|-------|
| <img width="342" height="611" alt="Image" src="https://github.com/user-attachments/assets/1e2ecf49-e69a-416a-9771-725cb5f1bec0" /> | <img width="342" height="611" alt="Image" src="https://github.com/user-attachments/assets/a53b8024-b071-4291-a320-2cae021839fc" /> |

### Version

Pre-release (Development)

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #495, opened 2025-12-22
Standardize club card font sizes (#494)

sy-pk:
### Changes
- Fixed club card title(20px) and body(16px) text to consistent sizes


TyHil:
This looks nice to me. @BK2004 what do you think?

Probably worth doing the same thing to the EventCard

Issue #453, opened 2025-12-10
Join club and Register event transitioning


TyHil:
### What did you expect to happen?

Clicking join club or event toggles the button state.

### What happened?

If the button is in state "Join", clicking it makes it show as loading the briefly flash as "Join" before going to "Joined". Same when going from "Joined" to "Join" or with the event register button. Why?

### Version

Pre-release (Development)

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #496, opened 2025-12-22
Fixed Join/Joined button flicker during state transitions

SaivishaalSureshkannan:
Fixes Join club and Register event transitioning #453 

Problem:
The Join/Joined button reverted to its previous label after loading for a brief time period. I believe this is happening due to the stale React query data during refetching. This causes the component to render again using that stale value, resulting in temporary reappearance of the initial label. 

Join -> Loading -> Join -> Joined
Joined -> Loading -> Joined -> Join

Solution:
I was able to implement an optimistic UI state for the join button. The button label updates immediately on click by the user and while that action is pending, the UI reflects the expected next state. If there is error, the UI will roll back. This removes that flicker. 

Testing:
- Verified Join -> joined and joined -> join transition on club homepage, specific club page, and the event section. 
- Used throttle to make

### DIFF
--- a/src/components/events/EventRegisterButton.tsx
+++ b/src/components/events/EventRegisterButton.tsx
@@ -5,6 +5,7 @@ import CheckIcon from '@mui/icons-material/Check';
 import { Button, Skeleton } from '@mui/material';
 import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query';
 import { useRouter } from 'next/navigation';
+import { useEffect, useState } from 'react';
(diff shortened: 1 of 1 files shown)


TyHil:
This works great! Can you do the same thing on the `src\components\events\EventRegisterButton.tsx`?

SaivishaalSureshkannan:
Hi, I was able to fix the button flicker on the register button on the event section as well.

SaivishaalSureshkannan:
Thanks for the review. I am ready to merge.

Issue #461, opened 2025-12-14
Event pagination


TyHil:
We have a lot of events for any org that imports their events using google calendar and has anything recurring which results in huge API call responses. Add pagination to the `/manage/[slug]/events` page as well as to the "From Your Joined Clubs" section of the Community page to reduce these response sizes.

PR #498, opened 2025-12-25
Limit upcoming events to 3 months and 20 items

sy-pk:
### Changes
- Limit upcoming events to the next 3 months with a maximum of 20 items


TyHil:
Can you make this a separate procedure for just the `ClubUpcomingEvents` component so it doesn't affect the `/manage/[slug]/events` and `/admin/clubs/[slug]/events` pages? We'll want those pages to still show all events (and later be paginated in #461)

PR #502, opened 2025-12-29
Timestamp for joining clubs and events

### LINKED ISSUE #500: Store timestamp that user joins a club or registers for an event
# Overview

When a user joins a club or registers for an event, the day and time this happened should be stored in `user_metadata_to_clubs` or `user_metadata_to_events`. 

The `MemberList` component from #490 should be modified to show this join time and to sort members by this join time by default

Isoscelestial:
Resolves #500 

Requires database migration

## Overview
- Adds `joinedAt` column to `user_metadata_to_clubs` table
- Adds `registeredAt` column to `user_metadata_to_events` table
- Show join/register date in tooltip of `JoinButton` and `EventRegisterButton`
- Show join date in club member list
  - Pro tip: Hold `ctrl + shift` to show the formatted locale time instead of the relative distance time. I directly copied this feature from the Discord server member page.
- Show join date for each club in account settings

### Join button tooltip

image

### Register button tooltip

image

### Club member list, with hover tooltip

The `Major` column is now hidden be default to make space for the new `Joined` column. It is still available via the columns button at the top right, though.

image

### Account settings, with join date on clubs

This page will be redesigned later, so I didn't put too much thought into how it should look.

image

### DIFF
--- a/src/components/manage/MemberList/MemberList.tsx
+++ b/src/components/manage/MemberList/MemberList.tsx
@@ -36,6 +36,7 @@ import useMemberListDeletionState from './useMemberListDeletionState';
 import {
   actionColumn,
   columns,
+  defaultUserSort,
   formatUserListString,
   MemberListAbilities,
   ToastState,
@@ -47,14 +48,12 @@ type MemberListProps = {
 };
 
 const MemberList = ({ members, club }: MemberListProps) => {
-  const membersIndexed = members
-    .sort((a, b) => b.joinedAt.getTime() - a.joinedAt.getTime())
-    .map((member, index) => {
-      return {
-        ...member,
-        id: index,
-      };
-    });
+  const membersIndexed = members.sort(defaultUserSort).map((member, index) => {
+    return {
+      ...member,
+      id: index,
+    };
+  });
 
   /**
    * Hooks and API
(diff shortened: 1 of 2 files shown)


src/components/manage/MemberList/CustomRenderCell.tsx
```diff
@@ -4,13 +4,79 @@ import VisibilityOutlinedIcon from '@mui/icons-material/VisibilityOutlined';
...
+      addSuffix: true,
+    },
+  );
+
+  return (
+    <Tooltip
+      title={localeDateString}
+      enterDelay={0}
+      placement="top"
+      arrow
```
TyHil: I think we could leave enterDelay, placement, arrow, and offset as default as we've done in other tooltips.

TyHil (approved):
This code looks great!

Unrelated, but we should totally do the "x months ago formatting" on the club listing and event listing pages for updated at and lasted event @AbhiramTadepalli.

src/components/manage/MemberList/CustomRenderCell.tsx
Isoscelestial: Perhaps, but I felt it's important here to include an arrow to remove any ambiguity about which row the tooltip is referring to. I think the arrowless tooltips for the remove and show/hide buttons are fine because they're the same on each row, but the join timestamp will be different on each row, so it's more important for the tooltip to clearly show which one it belongs to.

Although admittedly, I did sorta just mimic the implementation of Discord's server member list 😅

TyHil: Alright sounds good

TyHil:
Oh that's a misunderstanding on my part then. I'll merge this in and you can make that change in the refactoring PR.

PR #505, opened 2025-12-31
All tags feature request 

### LINKED ISSUE #418: All tags button
Add a small "All Tags" button on the homepage that allows adding any tag to the HomePageSearchBar. This way the user can explore any tag we have instead of just the top 7 or so below the search bar.

JoelZuniga:
Resolves #418 

Created a All Tags component allowing users to better filter through clubs in the search bar.

Also reduced the amount of the top tags shown under search bar to 5 to improve layout with component.

### DIFF
--- a/src/components/AllTags.tsx
+++ b/src/components/AllTags.tsx
@@ -48,7 +48,7 @@ export const AllTags = ({
     setShouldFocus(true);
   };
 
-  const sorted = [...new Set(options)].sort((a, b) => a.localeCompare(b));
+  const sorted = options.sort((a, b) => a.localeCompare(b));
 
   return (
     <>
@@ -66,8 +66,10 @@ export const AllTags = ({
         anchorEl={anchorEl}
         open={open}
         onClose={handleClose}
-        PaperProps={{ className: 'max-h-40 mt-2 rounded-xl' }}
-        MenuListProps={{ onClick: (e) => e.stopPropagation() }}
+        slotProps={{
+          paper: { className: 'max-h-40 mt-2 rounded-xl' },
+          list: { onClick: (e: React.MouseEvent) => e.stopPropagation() },
+        }}
       >
         {sorted.map((tag) => (
           <MenuItem key={tag} onClick={() => handleChange(tag)}>


JoelZuniga:
Currently the all tags component is sitting under the tags and its using the default outlined mui style, i wasn't sure if the look we were going for was for something more similar to the Tag Pills or something slightly different so i thought slightly different might look best.

src/components/AllTags.tsx
```diff
@@ -0,0 +1,74 @@
...
+    setTags(next);
+  };
+
+  const handleClose = () => {
+    if (!didChangeRef.current) return;
+    didChangeRef.current = false;
+
+    setShouldFocus(false);
+    scrollToResults();
+    setShouldFocus(true);
```
TyHil: I don't think we need to do both `setShouldFocus(false)` and `setShouldFocus(true)`. Just `setShouldFocus(true)`. Also seems `TagPill` doesn't call `setShouldFocus(true)`, would you mind adding that?

TyHil (changes requested):
Looking really good! Just one technical change and a design comment.

I'm thinking a design closer to the TagPills would be nicer. Like a more icon like this:
image
I make that mockup with this Button but it may not be possible to open the Select with a Button
```
<Button
  variant="contained"
  disableElevation
  size="small"
  className="drop-shadow-[0_0_4px_rgb(0_0_0_/_0.4)]"
>
  <MoreHorizIcon />
</Button>
```

src/components/AllTags.tsx
JoelZuniga: That design does look much better, Ill get that done. The design example is very useful, thank you.

src/components/AllTags.tsx
```diff
@@ -0,0 +1,80 @@
...
+      >
+        {label}
+      </Button>
+
+      <Menu
+        anchorEl={anchorEl}
+        open={open}
+        onClose={handleClose}
+        PaperProps={{ className: 'max-h-40 mt-2 rounded-xl' }}
+        MenuListProps={{ onClick: (e) => e.stopPropagation() }}
```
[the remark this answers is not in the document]
JoelZuniga: I used stopPropogation in order to prevent the Menu from closing after selecting just one tag allowing for multiple selections and only triggering the handleClose after the user clicks off of the menu or presses esc 

Ill replace the depreciated code, thanks for the suggestions!

Issue #456, opened 2025-12-13
User metadata overhaul


TyHil:
Currently user metadata is only able to be changed in a user's settings page and what we collect could be more helpful for our goal.

Update the user metadata to be more userful:
- Ask for graduation month/year instead of freshman/sophomore/...
- Ask for UTD email
- etc.

On a new user making an account, prompt them to fill in their user metadata.

Issue #406, opened 2025-12-03
Settings Update (Design)


BK2004:
Currently, the settings page does not match our other pages and just generally doesn't look good. Keeping the same fields, update the design to match our other pages.

Current design:

<img width="705" height="506" alt="Image" src="https://github.com/user-attachments/assets/ec0cf2b7-b067-48f1-8446-4ae290b882ed" />

PR #517, opened 2026-01-12
Redesign settings and add onboarding experience

### LINKED ISSUE #432: Redesign settings page
Waiting on #406 

The user settings page (`src\app\settings\page.tsx`) is outdated. It should use MUI components instead and have a design similar to `manage/[clubId]`.

You might want to abstract the form components from `app/manage/[clubId]` and put them in a new folder called `components/form`, so they can be used throughout the entire app when needed (for instance #406 )

Migrate from `react-hook-form` to `@tanstack/react-form` in the process.

Isoscelestial:
- Resolves #432, #456, #406
- Requires database migration

Let's see if I remember everything I put in this PR

## Overview

- Redesigned `/settings` page to more closely resemble the design by @anilanarula in the [Figma](https://www.figma.com/design/snmEoRUgfjAZYSc6hqySkY/Wireframes-for-Jupiter?node-id=6564-3356&t=dbLSlhcDsZmSI5UG-4)
  - I did, however, make a few tweaks to make it more consistent with the rest of the site
- Added onboarding experience for new users located in `/get-started`
  - When creating an account, users will automatically be redirected here
  - Uses fancy sliding multi-step form interface. If needed, I can abstract this stuff in the future so it can be used to redesign the club match form.
- Began abstracting the following form components into `components/form`:
  - `FormTextField`
  - `FormSelect`
  - `FormAutocomplete` (still work in progress, I probably overcomplicated things)
  - `FormFieldSet`
  - `FormQuestion`
- Added `student_classification` database enum, which can be "Student", "Graduate Student", "Alum", or "Prospective Student"
- Altered `user_metadata` table:
  - Added `student_classification`
  - Added `graduation_date`
  - Added `contact_email`
  - Dropped `year`
    - The data from this column is migrated to the new `student_classification` and `graduation_date` columns
- Added list of majors and minors to `constants/utdDegrees.ts`
  - Scraped majors from [here](https://academics.utdallas.edu/degrees)
  - Scraped minors from [here](https://catalog.utdallas.edu/current/undergraduate/minors)

## Notes

- The onboarding experience remains accessible to users even after they complete it; all they'd have to do is navigate to `/get-started`. If this is undesirable, we should add another column to the user metadata table called `completed_onboarding` and use that to prevent users from reopening the onboarding experience.
- The code for the onboarding experience is slightly janky, it probably creates a lot of unnecessary re-renders. I would appreciate if someone better than me at React would take a look at that
- I just realized I forgot to drop the year database enumeration. If someone sees this before I do that, please do that and I will be eternally grateful (I am not doing that right now because it is 4 am)
- Yes, everything is mobile friendly. Tested on responsive design mode in Firefox and Chrome, and iOS
- While testing this, I realized that when creating a new account, its `id` isn't a UUID. I don't know if this is because of a change I made or the database I have, but it shouldn't matter anyway since I removed anything that would break because of this.
- Although #432 and #456 are separate issues and I should have split this into two separate PRs, I felt they had very similar features and it would just be easier to get it done in a single PR. Sorry for that.
- Please check the custom migration file I wrote and make sure it works, since this is genuinely my first time ever writing SQL.

### Redesigned Settings (`/settings`)

account settings

### Onboarding Experience (`/get-started`)

https://github.com/user-attachments/assets/a41433df-5eca-4e3b-96fa-b22534b6aba1

### DIFF
--- a/src/components/getting-started/OnboardingFormStep.tsx
+++ b/src/components/getting-started/OnboardingFormStep.tsx
@@ -88,7 +88,7 @@ const OnboardingFormStep = withForm({
                         onChange={(value) => {
                           field.handleChange(value);
                         }}
-                        value={field.state.value}
+                        value={field.state.value ?? null}
                         label="Graduation Date"
                         className="[&>.MuiPickersInputBase-root]:bg-white dark:[&>.MuiPickersInputBase-root]:bg-neutral-900 w-64 grow"
                         slotProps={{


src/components/settings/SettingsForm.tsx
```diff
@@ -15,6 +18,7 @@ async function SettingsForm({
 
   const userData = await db.query.userMetadata.findFirst({
```
TyHil: Can you update this to use `api.userMetadata.byId`?

src/components/settings/SettingsForm.tsx
```diff
@@ -15,6 +18,7 @@ async function SettingsForm({
...
     where: eq(userMetadata.id, user.id),
+    with: { clubs: true },
   });
   if (!userData) return null;
   const joinedClubs = await db.query.userMetadataToClubs.findMany({
```
TyHil: Is joinedClubs needed in a separate query since `with: { clubs: true },` was added?

src/components/settings/SettingsHeader.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export default function SettingsHeader({ user }: SettingsHeaderProps) {
+  return (
+    <Panel className="bg-linear-to-r from-[#5A49F7] from-[4.36%] via-[#9403D8] via-[49.74%] to-[#FD9365]">
+      <div className="flex gap-4 max-sm:flex-col-reverse sm:flex-row pl-2">
+        <div className="flex flex-col gap-2">
+          <h1 className="font-display text-4xl font-semibold max-sm:text-center text-white">
+            Settings
+          </h1>
+          <span className="text-xl max-sm:text-lg max-sm:text-center text-white opacity-80">
```
TyHil: Can we put a text shadow on these, maybe `text-shadow-[0_0_16px_rgb(0_0_0_/_0.4)]` and `text-shadow-[0_0_4px_rgb(0_0_0_/_0.4)]`, from the homepage

src/server/db/schema/users.ts
```diff
@@ -26,6 +27,13 @@ export const roleEnum = pgEnum('role', [
...
 ]);
 
+export const studentClassificationEnum = pgEnum('student_classification', [
+  'Student',
+  'Graduate Student',
+  'Alum',
+  'Prospective Student',
+]);
+
 export const careerEnum = pgEnum('career', [
```
TyHil: I think we can drop career.

src/server/db/schema/users.ts
```diff
@@ -26,6 +27,13 @@ export const roleEnum = pgEnum('role', [
   'Administrator',
 ]);
 
+export const studentClassificationEnum = pgEnum('student_classification', [
+  'Student',
+  'Graduate Student',
+  'Alum',
+  'Prospective Student',
```
TyHil: Should we add Employee?

src/components/getting-started/OnboardingForm.tsx
```diff
@@ -0,0 +1,352 @@
...
+      Back
+    </Button>
+  );
+
+  const NextButton = (
+    <Button
+      type={activeStep.current === steps.length - 2 ? 'submit' : undefined}
+      variant="contained"
+      className="normal-case"
+      disabled={activeStep.current === steps.length - 2 && !form.state.isValid}
```
TyHil: Would it be better to just not let the user continue ever if the form is not valid? As in just `disabled={!form.state.isValid}`

src/components/getting-started/OnboardingForm.tsx
```diff
@@ -0,0 +1,352 @@
...
+                </Step>
+              ))}
+          </Stepper>
+        </div>
+      </BaseCard>
+      <Panel className="shadow-lg overflow-clip">
+        <div>
+          <div
+            className="relative max-sm:mb-2 sm:mb-4 transition-[height] duration-250 ease-in-out"
+            style={
```
TyHil: We can put this style prop on Panel instead of having these 2 divs I believe.

src/components/getting-started/OnboardingForm.tsx
```diff
@@ -0,0 +1,352 @@
...
+          >
+            {BackButton}
+            {NextButton}
+          </div>
+        </div>
+      </Panel>
+      <div
+        className={`${hideMobileStepper ? 'hidden' : 'max-sm:visible sm:hidden'}`}
+      >
+        <MobileStepper
```
TyHil: I like the mobile stepper but I think it'll get covered by the bug report button. Can we just keep the buttons in the card?

src/components/settings/forms/JoinedClubs.tsx
```diff
@@ -0,0 +1,190 @@
...
+      <div className="flex items-center gap-2 grow justify-end">
+        {canManage && (
+          <Chip
+            className="h-fit rounded-full bg-white border-1 border-slate-400 [&>.MuiChip-label]:inline-block [&>.MuiChip-label]:p-1.5"
+            label={
+              <div className="flex gap-2">
+                <MemberRoleChip memberType={joinedClub.memberType} />
+                <Button
+                  LinkComponent={Link}
+                  href={`/manage/${club.slug ?? club.id}`}
```
TyHil: Can you use a Link wrapping the Button. Its better for preloading.

Also slug should always be defined, so no need for ?? club.id

src/components/getting-started/FormStep.tsx
```diff
@@ -0,0 +1,224 @@
...
+      <div id={active ? 'active-form-step' : undefined} className="mx-2">
+        {FormStepData}
+      </div>
+    );
+  },
+});
+
+export default FormStep;
+
+function FormStepDataPanel({
```
TyHil: Ok wait, maybe the Question is all good but Panel could be used for this...

TyHil (changes requested):
Ok! Looking really good! Here's the review

src/components/getting-started/OnboardingForm.tsx
Isoscelestial: I had that originally, but I was having a problem where because validation only runs once the user changes an input, a user could still skip ahead even if it wasn't valid. I resolved this by... making the graduation date optional. I'll look into a more robust solution

src/components/settings/forms/JoinedClubs.tsx
Isoscelestial: I found that a Link wrapping the button affects keyboard accessibility by causing there to be two items that are tab-able: the link and the button. It also breaks the functionality of drag-and-dropping the link someplace, which tbf not really a commonly used feature but still something that breaks nonetheless.

Nevertheless, I'll try to find a way wrap the button with Link and find a way to resolve these minor accessibility bugs

src/components/form/FormQuestion.tsx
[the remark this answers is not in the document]
Isoscelestial: On the second step of onboarding that asks for college info, the description/question part is needed twice (to break up the inputs). That's why I made this question component: to contain question text plus response inputs. It's a little ambiguous that the question text and future Panel description are styled identically, but I feel like it's fine to have flexibility as it really depends on how the forms throughout the website are structured.

src/components/getting-started/OnboardingForm.tsx
TyHil: I'm not sure what you mean by statically and underneath everything, could you help me out?

I agree that the stepper won't look good on mobile. I don't get why we need the back/next buttons at the bottom of the page instead of in the same place as on desktop tho. Maybe to indicate what step number the user is on we could have just the current step at the top of the page instead of the whole stepper

src/components/settings/forms/JoinedClubs.tsx
Isoscelestial: Yeah, the way you did it was what documentation online suggested, but it seems like the MUI Button `component` prop was replaced by a `LinkComponent` prop so I used that instead. It seems to work fine, although I'm not sure how to test how well it preloads as you mentioned in your original comment

TyHil: I am somewhat confident that will do the preloading, though it's hard to test when it's not on prod. We can do that here and see how it works once its on prod, then make an issue to convert the rest.

src/components/settings/SettingsForm.tsx
TyHil: Hmm maybe switch to byId and replace joinedClubs with a new procedure that makes the same query as joinedClubs, since I know we'll need that procedure later for the my community page changes.

So still 2 calls here, just both are procedures instead of db.query

src/components/settings/forms/JoinedClubs.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+type ClubListItemProps = ClubListItemPropsBase &
+  (
+    | {
+        joinedClub: SelectUserMetadataToClubsWithClub;
+        club?: never;
+      }
+    | {
+        joinedClub?: never;
+        club: SelectClub;
```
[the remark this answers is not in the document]
TyHil: Yeah, this component is pretty specific to JoinedClubs, it seems like unnecessary added complexity.

src/components/getting-started/OnboardingForm.tsx
Isoscelestial: I made it so that if you try to navigate to the next/previous step, it'll stop you if there are errors on that step. Ideally, it wouldn't prevent you from going backward, but that would've complicated things with the next button being disabled, so... this will work.

TyHil: Sounds great!

src/server/db/schema/users.ts
Isoscelestial: If we did, we should use the terminology "Faculty" and/or "Staff" (either as two new items or a single item called "Faculty/Staff"). Might be a good idea, which one(s) should I add?

TyHil: 2 new items sounds great!

Isoscelestial:
Things to note:
- Migrations for removing the unused enums and columns for user metadata have not been generated since we're currently facing issues with that
- Although you didn't request this, I somewhat refactored how the multi-step form thing works because I intend on further abstracting this into its own component in `src/components/form/FormWizard.tsx` at some point in the future, as I believe it may be useful in case we wanted a sliding form wizard UI elsewhere (club match? club creation? who knows)
- Also added dark mode

TyHil (approved):
Sweet!

PR #528, opened 2026-01-15
Contact card fixes

AbhiramTadepalli:
Tried to make most text fit. Some will still inevitably exceed the width on desktop (not mobile tho)
Ensured that insta handles don't have trailing stuff


TyHil (changes requested):
This is great! One thing, can you add the same query param removal to twitter, facebook, youtube, and twitch?

PR #535, opened 2026-01-16
Event page redesign

TyHil:
I did not implement any of the new feature from the design but added all of the ones that we currently had the capability to do. We can discuss things like register by dates, paid registration, event tags, multiple host clubs, and location room linking in meetings.

https://github.com/user-attachments/assets/160e0151-f7fc-4078-9241-58debae9e810

### DIFF
--- a/src/components/events/EventForm.tsx
+++ b/src/components/events/EventForm.tsx
@@ -14,7 +14,10 @@ import { type SelectClub } from '@src/server/db/models';
 import { useTRPC } from '@src/trpc/react';
 import { type RouterOutputs } from '@src/trpc/shared';
 import { useAppForm } from '@src/utils/form';
-import { createEventSchema, eventFormSchema } from '@src/utils/formSchemas';
+import {
+  createEventFormSchema,
+  editEventFormSchema,
+} from '@src/utils/formSchemas';
 import EventCard, { EventCardSkeleton } from './EventCard';
 
 type EventFormProps =
@@ -150,7 +153,7 @@ const EventForm = ({ mode = 'create', club, event }: EventFormProps) => {
       );
     },
     validators: {
-      onChange: mode === 'create' ? createEventSchema : eventFormSchema,
+      onChange: mode === 'create' ? createEventFormSchema : editEventFormSchema,
     },
   });
 
--- a/src/server/api/routers/event.ts
+++ b/src/server/api/routers/event.ts
@@ -23,7 +23,7 @@ import {
   userMetadataToEvents,
 } from '@src/server/db/schema/users';
 import { dateSchema, order } from '@src/utils/eventFilter';
-import { createEventSchema, updateEventSchema } from '@src/utils/formSchemas';
+import { createEventSchema, editEventSchema } from '@src/utils/formSchemas';
 import { getGoogleAccessToken } from '@src/utils/googleAuth';
 import { callStorageAPI } from '@src/utils/storage';
 import { createTRPCRouter, protectedProcedure, publicProcedure } from '../trpc';
@@ -370,7 +370,7 @@ export const eventRouter = createTRPCRouter({
       return newEvent.id;
     }),
   update: protectedProcedure
-    .input(updateEventSchema)
+    .input(editEventSchema)
     .mutation(async ({ input, ctx }) => {
       const { id, clubId, ...data } = input;
       const userId = ctx.session.user.id;
--- a/src/utils/formSchemas.ts
+++ b/src/utils/formSchemas.ts
@@ -158,7 +158,7 @@ export const editSlugSchema = z.object({
     ),
 });
 
-export const createEventSchema = z.object({
+export const createEventFormSchema = z.object({
   clubId: z.string(),
   name: z
     .string()
@@ -174,25 +174,29 @@ export const createEventSchema = z.object({
   image: fileSchema,
 });
 
-export const updateEventSchema = createEventSchema.extend({
+export const createEventSchema = createEventFormSchema.omit({
+  image: true,
+});
+
+export const editEventFormSchema = z.object({
+  clubId: z.string(),
+  name: z
+    .string()
+    .min(3, 'Name must be at least 3 characters')
+    .max(100, 'Character limit reached'),
+  location: z
+    .string()
+    .min(1, 'Location is required')
+    .max(100, 'Character limit reached'),
+  description: z.string().max(1000, 'Character limit reached'),
+  startTime: z.date('Invalid date'),
+  endTime: z.date('Invalid date'),
+  image: fileSchema,
+});
+
+export const editEventSchema = editEventFormSchema.extend({
   image: z.url().nullable(),
   id: z.string(),
-});
-
-export const eventFormSchema = z.object({
-  clubId: z.string(),
-  name: z
-    .string()
-    .min(3, 'Name must be at least 3 characters')
-    .max(100, 'Character limit reached'),
-  location: z
-    .string()
-    .min(1, 'Location is required')
-    .max(100, 'Character limit reached'),
-  description: z.string().max(1000, 'Character limit reached'),
-  startTime: z.date('Invalid date'),
-  endTime: z.date('Invalid date'),
-  image: fileSchema,
 });
 
 const characterLimitError = 'Character limit reached';


src/components/events/listing/EventTitle.tsx
```diff
@@ -0,0 +1,76 @@
...
+}: {
+  event: NonNullable<RouterOutputs['event']['getListingInfo']>;
+}) => {
+  const startTime = new TZDateMini(event.startTime, 'America/Chicago');
+  const endTime = new TZDateMini(event.endTime, 'America/Chicago');
+
+  return (
+    <section
+      id="event-title"
+      className="w-full rounded-lg flex flex-col md:flex-row items-start md:items-center justify-between gap-4 mb-0 mt-2"
```
Isoscelestial: In my opinion, this top margin makes the title look unbalanced, as the register buttons are no longer vertically centered between the banner and the body. I think remove this and update the `main` component in `/events/[id]/page.tsx` to have a gap of 8 instead of 6, so that everything has a bit more room to breath.

src/components/events/listing/EventDetailsCard.tsx
```diff
@@ -0,0 +1,48 @@
...
+  event: NonNullable<RouterOutputs['event']['getListingInfo']>;
+  id?: string;
+};
+export default function EventDetailsCard({ event, id }: EventDetailsCardProps) {
+  return (
+    <Panel className="text-sm" id={id} smallPadding heading="Details">
+      {event.numParticipants || event.updatedAt ? (
+        <>
+          {event.numParticipants !== 0 && (
+            <div className="flex flex-row flex-wrap gap-1 py-1 border-b border-slate-100 last:border-0">
```
Isoscelestial: Dark mode border styling

```suggestion
            <div className="flex flex-row flex-wrap gap-1 py-1 border-b border-slate-100 dark:border-slate-800 last:border-0">
```

src/components/events/EventRegisterButton.tsx
```diff
@@ -9,12 +9,20 @@ import { useEffect, useState } from 'react';
...
-type buttonProps = {
+type EventRegisterButtonProps = {
   isHeader?: boolean;
+  clubId: string;
+  clubSlug: string;
   eventId: string;
 };
-const EventRegisterButton = ({ isHeader, eventId }: buttonProps) => {
+const EventRegisterButton = ({
+  isHeader,
```
Isoscelestial: I think this prop should also control the order of the Edit and Register buttons:

- When true, the register button is on the right side (for the event page)
- When false, the register button is on the left side (for Event Card)

This way, the register button is always positioned at the same location, no matter whether the user has the ability to edit the event or not.

src/components/club/listing/ClubUpcomingEventsCard.tsx
```diff
@@ -3,13 +3,15 @@ import EventCard from '@src/components/events/EventCard';
...
 
 type ClubUpcomingEventsCardProps = {
-  club: NonNullable<RouterOutputs['club']['getDirectoryInfo']>;
+  emptyText: string;
+  heading: string;
   upcomingEvents: NonNullable<RouterOutputs['event']['byClubId']>;
   id?: string;
 };
 
 export default function ClubUpcomingEventsCard({
```
Isoscelestial: I feel like this component shouldn't show all of a club's other events when used on an event view page. Perhaps add a prop that will limit the list to only 3 other events and displays a banner underneath that says "More events on club page..." and links to the club page with a URL fragment of `#upcoming-events` (so it automatically scrolls down to the upcoming events section)

TyHil: Right now it shows the next 20 events within 3 months. It definitely adds to the scroll of the page but I dont think we need to make the user go to the clubs directory listing to see more

AbhiramTadepalli (changes requested):
- [x] can you make the host club image smaller on desktop
- [x] It's not immediately obvious you can click on the host club. Maybe add a card around the logo+name with a hover effect

AbhiramTadepalli (approved):
Lgtm

AbhiramTadepalli:
Nitpicking, but maybe slightly less border radius

src/components/events/listing/EventCountdownCard.tsx
```diff
@@ -0,0 +1,140 @@
...
+  id?: string;
+};
+
+export default function EventCountdownCard({
+  startTime,
+  endTime,
+  id,
+}: EventCountdownCardProps) {
+  const eventStartTime = startTime.getTime();
+  const eventEndTime = startTime.getTime();
```
Isoscelestial: Typo broke this

```suggestion
  const eventEndTime = endTime.getTime();
```

src/components/events/listing/EventCountdownCard.tsx
```diff
@@ -0,0 +1,140 @@
...
+    return <Panel id={id} smallPadding heading="Event Starts In" />;
+  }
+
+  if (now > eventStartTime) {
+    if (now < eventEndTime) {
+      return (
+        <Panel id={id} smallPadding className="text-center">
+          <p className="font-display text-2xl">Happening Now</p>
+          <p className="text-sm font-medium text-slate-600 dark:text-slate-400">
+            Ends {endsIn}
```
Isoscelestial: Can't you do the same thing you did from lines 101-104 with the "Ended ..." text? Then remove the unnecessary `endsIn` state, which I believe removes an unneceesary re-render.

```suggestion
            Ends{' '}
            {formatDistanceStrict(eventEndTime, now, {
              addSuffix: true,
            })}
```

src/components/events/listing/EventCountdownCard.tsx
```diff
@@ -0,0 +1,140 @@
...
+  );
+  const timeUntilStartInDays = Math.floor(
+    timeUntilStart / (1000 * 60 * 60 * 24),
+  );
+
+  return {
+    days: timeUntilStartInDays.toString(),
+    hours: timeUntilStartInHours.toString(),
+    minutes: timeUntilStartInMinutes.toString(),
+    seconds: timeUntilStartInSeconds.toString(),
```
Isoscelestial: I think the seconds should always have two digits

```suggestion
    seconds: timeUntilStartInSeconds.toString().padStart(2, '0'),
```

Isoscelestial: Pad the minutes too. However, I think it'd be cool to only have two digits if there's hours left. That way, when the count is at less than 10 minutes, it is easier to read how much time is left (since 9m 50s is easier to read than 09m 50s, yet 3h 04m 27s is easier to read than 3h 4m 27s, IMO).

```suggestion
    minutes: timeUntilStartInMinutes
      .toString()
      .padStart(timeUntilStartInHours > 0 ? 2 : 0, '0'),
```

src/components/events/listing/EventTitle.tsx
```diff
@@ -0,0 +1,92 @@
...
+        )}
+        <div className="flex flex-wrap flex-col lg:flex-row gap-4 text-slate-600 dark:text-slate-400 lg:items-center text-sm md:text-base">
+          <span className="flex gap-2 items-center">
+            <EventIcon className="text-xl md:text-2xl" />
+            {format(startTime, 'EEE, LLLL d, yyyy @ h:mm a')}
+          </span>
+          <Divider orientation="vertical" flexItem className="hidden lg:flex" />
+          <span className="flex gap-2 items-center">
+            <AccessTimeIcon className="text-xl md:text-2xl" />
+            {`Lasts ${formatDuration(
```
Isoscelestial: It is currently possible to set an event's end time to be the same as the start time. Doing this causes `formatDuration()` to output nothing, which results in "Lasts (till 12:34 AM)". I think we should have a backup string that shows up when `startTime === endTime`; maybe default to "Lasts 1m", since "Lasts 0 minutes" looks like sloppy programming and I honestly can't think of something else (besides flat out removing this event duration item, but I don't think we should do that because that creates potential inconsistency between events and may confuse users.)

Obviously, we could prevent events from having the same start and end time in the first place, but I think it'd be nice to natively support it in the front end anyway, just in case

Isoscelestial (changes requested):
Looks very nice! I have a few UI/UX opinions and just a single bug, and then I think this is ready to be merged

src/components/events/listing/EventTitle.tsx
TyHil: Maybe just "No end time specified"

I'll make an issue to enforce that the endTime is after the startTime in event creation.

src/components/events/listing/EventCountdownCard.tsx
TyHil: I don't know if we need this much complexity. It used to always be 2 digits, I can go back to that. Maybe it would be more readable to have the days disappear when they hit 0, same with the hours, then minutes.

src/components/club/listing/ClubEventHeader.tsx
```diff
@@ -9,18 +9,17 @@ type Club = SelectClub & {
...
   return (
     <BaseCard className="relative w-full aspect-[4.5/1] overflow-hidden">
       <Image
         src={club.bannerImage ?? club.profileImage ?? '/images/wideWave.jpg'}
         alt="Club banner"
         fill
         className="object-cover object-center"
-        priority
       />
     </BaseCard>
```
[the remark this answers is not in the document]
TyHil: Fair point. Do you think the profile image should be used as a backup banner or just not show it until there's an actual banner?

src/components/events/listing/EventHostClubCard.tsx
```diff
@@ -0,0 +1,37 @@
...
+  club: NonNullable<RouterOutputs['event']['getListingInfo']>['club'];
+  id?: string;
+};
+export default function EventHostClubCard({
+  club,
+  id,
+}: EventHostClubCardProps) {
+  return (
+    <Panel className="text-sm" id={id} smallPadding heading="Host Club">
+      <Link
```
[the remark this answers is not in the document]
TyHil: I'm not a big fan of "Click to", how about just "View club directory page". I could make the same change on settings

src/components/events/listing/EventCountdownCard.tsx
Isoscelestial: Huh, when I removed the `endsIn` state it still worked fine for that case. I wonder what was triggering the re-renders. But that makes sense

src/components/events/listing/EventCountdownCard.tsx
```diff
@@ -0,0 +1,187 @@
+'use client';
+
+import { time } from 'console';
```
Isoscelestial: Unused import

PR #536, opened 2026-01-17
Fix: attempting to delete invalid IDs

TyHil:
Resolves https://utdnebula.sentry.io/issues/7113592710/

New officers are given an ID `new-${currentOfficers.length}` but then when they're removed they're added to the list of IDs to delete because they have an ID. This adds a check so only IDs that don't begin with "new" are added.


Isoscelestial (dismissed):
Good catch, looks good to me. Just gotta remove the unused import for `useEffect`.

PR #558, opened 2026-01-28
Search users by contact email Search users by contact email

SinhSinhAn:
Adds UTD contact email to the user search backend so searches match full name, account email, or UTD email. This improves UserSearchBar suggestions for users whose UTD email is the primary identifier.


TyHil (approved):
Awesome! Thank you for this!

PR #562, opened 2026-01-29
Use follow wording for clubs

SinhSinhAn:
Change Join to Follow  issue#556 fixed

### DIFF
--- a/src/components/manage/MemberList/CustomRenderCell.tsx
+++ b/src/components/manage/MemberList/CustomRenderCell.tsx
@@ -131,11 +131,7 @@ const RoleToMemberType: Record<string, MemberTypes> = {
 export function MemberTypeCell(params: GridRenderCellParams) {
   if (!params.value) return;
   return (
-    <MemberRoleChip
-      memberType={
-        RoleToMemberType[params.value] ?? RoleToMemberType.Follower ?? 'Member'
-      }
-    />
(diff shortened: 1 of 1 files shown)


Realistically1:
Double checked members to followers across multiple sources, ready for reviews

src/components/manage/MemberList/CustomRenderCell.tsx
```diff
@@ -125,13 +125,17 @@ export function ContactEmailCell(params: GridRenderCellParams) {
...
 
 export function MemberTypeCell(params: GridRenderCellParams) {
   if (!params.value) return;
   return (
-    <MemberRoleChip memberType={RoleToMemberType[params.value] ?? 'Member'} />
+    <MemberRoleChip
+      memberType={
+        RoleToMemberType[params.value] ?? RoleToMemberType.Follower ?? 'Member'
+      }
+    />
```
TyHil: I misunderstood how this code works. I don't think this change is necessary. Since `RoleToMemberType.Follower` is always defined so the `?? Member` isn't used. So you can just revert this

Realistically1:
Reverted fallback as per review, logic already covers it

TyHil (approved):
Sweet, thank you!

Realistically1:
Yayyyyyyy!

PR #541, opened 2026-01-21
Calendar sync

AbhiramTadepalli:
closes https://github.com/UTDNebula/utd-clubs/issues/463, https://github.com/UTDNebula/utd-clubs/issues/471,

PR #563, opened 2026-01-29
Fix null refresh tokens by forcing re-login on home page

AbhiramTadepalli:
Building off of #541, this PR detects a null refresh token on the home screen and prompts consent again to properly catch a refresh_token.

Works on all pages


nl32 (approved):
There may be a way to do this server side by hooking into the better-auth server but this works and I don't see a clear path to a better solution.

PR #565, opened 2026-01-29
Convert button links to LinkComponent

SinhSinhAn:
Refactors MUI buttons that were wrapped in Next.js Links to use LinkComponent + href instead, improving accessibility and native link behaviors


src/components/header/ClubMatchButton.tsx
TyHil: Would you use the same changes from this branch instead of this refactor: https://github.com/UTDNebula/utd-clubs/blob/feature/tweaked-header-and-logo/src/components/header/ClubMatchButton.tsx

Just to avoid conflicts.

TyHil (changes requested):
Can you update the `RedoClubMatchButton` to use the `LinkComponent` prop instead of `component`.

Then checkout develop, pull changes, checkout your branch, and merge develop into your branch and resolve the conflict. Or just resolve the conflict online with the button at the bottom of the page.

src/components/header/ClubMatchButton.tsx
Realistically1: I could not find this, did you resolve it already? I get a 404 when I click on it

PR #576, opened 2026-01-31
Validate event end time after start

SinhSinhAn:
Adds Zod refinements to event create/edit forms to ensure end time is after start time and surfaces a field error on endTime, lmk if there's any issue with formatting, if need I can add some comments for better ... readability?


Realistically1:
wait hang on running into some errors, lemme see what's happening

TyHil (dismissed):
Tested and looks good!

TyHil:
Wait actually one thing, can you remove the periods from the ends of the error messages? They end up with a double period when rendered since periods are also added later.

src/utils/formSchemas.ts
TyHil: Missed one

PR #626, opened 2026-02-15
Fix all-day GCal events displaying with wrong times issue 607 

### LINKED ISSUE #607: Handle Gcal all-day events
# Overview

Currently, all-day events in gcal run from like 6:00pm the previous day until 5:59pm the next day after the event, effectively increasing its length by 24 hours.

# Proposal

When importing a full-day event from Gcal, make sure it runs from 12am-11:59pm of that day.

## Supplementary Information

(A) Basically need to convert a full day event into a 24hr event instead.
(B) @TyHil @BK2004 do we want to display full-day events in a different way?

SinhSinhAn:
Append T00:00:00 to date-only strings so TZDateMini interprets them as midnight in America/Chicago instead of midnight UTC.  Also remove the extra addDays(1) on the end time since Google's end.date is already exclusive.

### DIFF
--- a/src/utils/calendar.ts
+++ b/src/utils/calendar.ts
@@ -228,13 +228,13 @@ function generateEvent(
     recurrence: JSON.stringify(event.recurrence),
     recurenceId: event.recurringEventId,
     startTime: event.start.date
-      ? new TZDateMini(event.start.date, 'America/Chicago')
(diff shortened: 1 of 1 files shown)


SinhSinhAn:
Awesome, I'll note this down for future uses , please let me know if there's anything else you would like me to change

AbhiramTadepalli (approved):
lgtm

BK2004 (approved):

PR #630, opened 2026-02-18
Feature/improved-panels

Isoscelestial:
wip I'll add stuff here later


Isoscelestial:
Several of the features in this PR were merged with #651, notably:
- collapsible panels
- transparent panel variant
- updated collapsible panel for club listing's officer list

Closing this PR because it uses client-side code for creating the table of contents, which is absolutely not the best way to do things. Plus, we plan on cleaning up the management pages so there won't be as many panels, thereby making a table of contents feature less important.

PR #634, opened 2026-02-18
Add Microsoft to BetterAuth for Staff Sign in

### LINKED ISSUE #485: Add email sign ins
# Overview
Allow users to sign in with an email & password (could restrict to @utdallas.edu emails only perhaps)

justinschwerdtfeger:
This fixes #485

> [!IMPORTANT]
> This commit adds two new required environment variables , these are
> MICROSOFT_CLIENT_ID and MICROSOFT_CLIENT_SECRET

Right now, the UI looks like this. Clicking "Sign in here" starts BetterAuth sign in with Microsoft. 
image

We could potentially reject non-utd emails, but I'm not sure if that's a good idea. Right now, any Microsoft email will work.

There's two reasons why this PR doesn't just add "Sign in with Microsoft Button":
1. UTD deletes Microsoft accounts after graduation. This can cause issues with transferring club ownership post-graduation
2. Most users have a personal Microsoft account, a UTD account, a Gmail, and a Discord account. Having so many options can make it easy to forget which account the user signed into clubs with, creating duplicate accounts, and making it harder for club managers


TyHil:
I added the prod environment variables to Vercel. They won't work on the Vercel preview deployment tho because I can't use wild cards in the Microsoft redirect URL so only `https://clubs.utdnebula.com/api/auth/callback/microsoft` and `https://dev.clubs.utdnebula.com/api/auth/callback/microsoft` are configured. Would you dm the dev credentials to Braedon, Ethan, and Abhiram so they can test locally?

src/server/auth.ts
```diff
@@ -25,6 +25,10 @@ export const auth = betterAuth({
       clientId: env.DISCORD_CLIENT_ID,
       clientSecret: env.DISCORD_CLIENT_SECRET,
     },
+    microsoft: {
+      clientId: env.MICROSOFT_CLIENT_ID,
```
TyHil: Could you add a comment here about how these will expire on 2028-02-11

justinschwerdtfeger: I believe you can also set up azure to email on certificate expiration

BK2004 (changes requested):
Looks good! Just need to update the .env.example file with client ID and secret.

justinschwerdtfeger:
Good idea

BK2004 (approved):

PR #647, opened 2026-03-03
Update login redirect behavior so sers land on /community after authe…

### LINKED ISSUE #632: For logged-in users, redirect to the `/my-community` page on first load
# Overview

Logged-in users probably want to see an overview of their clubs and events (`/my-community`), rather than a big search box

# Proposal

When a logged-in user visits `clubs.utdnebula.com` for the first time (per session), redirect to the community page.

## Options

1. use a session storage flag indicating if it's the user's first time visiting the home page that session
2. make the community page default, and reroute logged-out users to the search page

Minhnimum:
Updated login redirect behavior so users land on /community after authentication, instead of /get-started.

Fixed hydration mixmatch in header profile/session UI by adding mount-safe rendering guards.

Fixes #632

### DIFF
--- a/src/app/auth/ProviderButtons.tsx
+++ b/src/app/auth/ProviderButtons.tsx
@@ -23,8 +23,8 @@ const ProviderButton = ({
     onClick={() => {
       void authClient.signIn.social({
         provider: provider,
-        callbackURL: callbackUrl ?? window.location.href,
-        newUserCallbackURL: '/get-started',
+        callbackURL: callbackUrl ?? '/community',
+        newUserCallbackURL: '/community',
       });
     }}
     className="bg-white hover:bg-neutral-200 dark:bg-neutral-800 dark:hover:bg-neutral-700 normal-case whitespace-nowrap min-w-max"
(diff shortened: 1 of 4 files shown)


TyHil:
I believe this works only one time when an existing user logs in whereas the proposal was:

> When a logged-in user visits `clubs.utdnebula.com` for the first time (per session), redirect to the community page.

where session I'd guess referred to `sessionStorage` which is unique to a tab. So we'd want to check something like `if (sessionStorage.getItem('newTab') == null)` then `sessionStorage.setItem('newTab', false)` and redirect.

Also, the hydration mixmatch fix is causing the linting check to fail. Run `npm run lint` to see why but that would probably be a change better suited for a separate PR.

Isoscelestial:
@Minhnimum Hey, just checking in to see if you need any help with this PR? We have our [last meeting](https://clubs.utdnebula.com/events/386ohbjibrfi6da33if49nb2g9_20260428T223000Z) at 5:30 today in SCI 3.240 if you want to ask us any questions

PR #649, opened 2026-03-05
fixed issue #644 "Can't leave club I am an admin/collaborator of from settings page"

### LINKED ISSUE #644: Can't leave club I am and admin/collaborator of from the settings page
### What happened?

The "trash" or unfollow button appears for clubs I am and admin/collabroator for:

Image

However, due to the member check in the `joinLeave` procedure in `src\server\api\routers\club.ts`, clicking the button just produces and error.

### What to do?

Update the button to work the same as the button in `src\components\manage\CollaboratorListItem.tsx` which disables the button when a user is a collaborator/admin and has the below Tooltip:

```
<Tooltip
  title={
    canTogglePresident
      ? self && (
          <div className="text-center">
            You cannot take away your own admin status
            <br />
            Another admin must remove you
          </div>
        )
      : 'Only an admin can change admin status'
  }
>
```

### Version

clubs.utdnebula.com

_No response_

### Relevant log output

```shell

```

### Code of Conduct

AliBenrami:
Added a warning message and disabled the unfollow button when the user is an admin or collaborator of a club

### DIFF
--- a/src/components/settings/forms/JoinedClubs.tsx
+++ b/src/components/settings/forms/JoinedClubs.tsx
@@ -174,10 +174,29 @@ function ClubListItem({ joinedClub, onLeave }: ClubListItemProps) {
             }
           />
         )}
-        <Tooltip title="Unfollow club">
-          <IconButton aria-label="unfollow" onClick={onLeave}>
-            <DeleteIcon />
-          </IconButton>
+        <Tooltip
+          title={
+            canManage
+              ? self && (
+                  <div className="text-center">
+                    You cannot Unfollow this club
+                    <br />
+                    Another admin must remove you
+                  </div>
+                )
+              : 'Unfollow club'
+          }
+        >
+          {/* This span is required to ensure the locked tooltip shows when the IconButton is disabled */}
+          <span>
+            <IconButton
+              aria-label="unfollow"
+              onClick={onLeave}
+              disabled={canManage}
+            >
+              <DeleteIcon />
+            </IconButton>
+          </span>
         </Tooltip>
       </div>
     </div>
--- a/src/components/settings/forms/JoinedClubs.tsx
+++ b/src/components/settings/forms/JoinedClubs.tsx
@@ -83,7 +83,17 @@ export default function JoinedClubs({ joinedClubs }: ClubsProps) {
                 );
 
                 joinedClubs.splice(removeIndex, 1);
+                setSnackbar(
+                  SnackbarPresets.savedCustom('Left club!'),
+                );
               },
+              onError: (e) => {
+                setSnackbar(
+                  SnackbarPresets.errorCustomMessage(
+                    'Failed to leave club!',
+                    e.message,
+                  ),
+                );
             },
           );
         }}
--- a/src/components/settings/forms/JoinedClubs.tsx
+++ b/src/components/settings/forms/JoinedClubs.tsx
@@ -14,6 +14,7 @@ import Link from 'next/link';
 import { useState } from 'react';
 import Panel from '@src/components/common/Panel';
 import Confirmation from '@src/components/Confirmation';
+import { setSnackbar, SnackbarPresets } from '@src/components/global/Snackbar';
 import MemberRoleChip from '@src/components/manage/MemberRoleChip';
 import { SelectUserMetadataToClubsWithClub } from '@src/server/db/models';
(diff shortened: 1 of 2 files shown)


src/components/settings/forms/JoinedClubs.tsx
```diff
@@ -174,10 +174,29 @@ function ClubListItem({ joinedClub, onLeave }: ClubListItemProps) {
...
           />
         )}
-        <Tooltip title="Unfollow club">
-          <IconButton aria-label="unfollow" onClick={onLeave}>
-            <DeleteIcon />
-          </IconButton>
+        <Tooltip
+          title={
+            canManage
+              ? self && (
```
TyHil: Idt `self` is defined is defined here. Idk how this built. Can you just remove `self`  and just do `canManage ? ... : ...`

TyHil (changes requested):
One quick change and this'll be good!

TyHil (dismissed):
Nice!

src/components/settings/forms/JoinedClubs.tsx
```diff
@@ -174,10 +174,29 @@ function ClubListItem({ joinedClub, onLeave }: ClubListItemProps) {
...
         )}
-        <Tooltip title="Unfollow club">
-          <IconButton aria-label="unfollow" onClick={onLeave}>
-            <DeleteIcon />
-          </IconButton>
+        <Tooltip
+          title={
+            canManage ? (
+              <div className="text-center">
+                You cannot Unfollow this club
```
Isoscelestial: I think we should use clearer language that explains why you can't leave, rather than just stating "oh you can't leave haha"
```suggestion
                You cannot unfollow a club you manage
```

Isoscelestial (commented):
Looks good, though I'd change the language to be clearer.

I do have a concern, though. In some places, we prohibit collaborators and admins from leaving (e.g. here with the `joinLeave` club procedure), but in other places we only prohibit admins from leaving (e.g. the `removeMembers` clubEdit procedure. We should make this consistent.

I believe the original intent of this restriction is to prevent cases where there are clubs with no admins in it. That's why I think it's fine to let collaborators leave, but not admins. Thoughts on what we should do?

TyHil:
Yeah I'd agree @Isoscelestial. @AliBenrami if you wanna work it into this PR to let collaborators leave but not admins in those listed places that'd make sense

AliBenrami:
How are we storing admin permissions? From what I can tell, we are only checking whether a user can leave a club based on their member type, but I do not see any other place where admin status is being referenced. Any wisdom would be appreciated.

TyHil:
An admin is listed as a "President" in the db, a collaborator is listed as an "Officer", and a follow is listed as a "Member". That's all in the `userMetadataToClubs` table defined in `src/server/db/schema/users.ts`

src/components/settings/forms/JoinedClubs.tsx
Isoscelestial, line 88: I think it might be nice to add snackbar feedback to show success and errors to the user, which is consistent with JoinButton.

```suggestion
                setSnackbar(
                  SnackbarPresets.savedCustom('Left club!'),
                );
              },
              onError: (e) => {
                setSnackbar(
                  SnackbarPresets.errorCustomMessage(
                    'Failed to leave club!',
                    e.message,
                  ),
                );
```

Isoscelestial (changes requested):
Looking good. Just gotta change the backend to allow officers to leave.

Update lines 291-299 of `src/server/api/routers/club.ts` to:

```ts
      if (dataExists) {
        if (dataExists.memberType !== 'President') {
          await ctx.db
            .delete(userMetadataToClubs)
            .where(
              and(
                eq(userMetadataToClubs.userId, joinUserId),
                eq(userMetadataToClubs.clubId, clubId),
              ),
            );
        } else {
          throw new TRPCError({
            code: 'FORBIDDEN',
            message: 'Cannot remove yourself because you are an admin',
          });
        }
```

I looked everywhere else and I think this is the only instance of collaborators not being able to leave. We do prevent people from removing themselves in the club management pages, but I think that's fine since it prevents people from accidentally locking themselves out of their club.

AliBenrami:
@Isoscelestial thx for the help!

Isoscelestial (approved):
Looks good to me!

BK2004 (approved):
Looks good!

PR #650, opened 2026-03-09
Redesigned events directory

Isoscelestial:
Moved to #651 

## Overview
- Redesigns `/events` page
- Creates filter components for use in UTD Trends as well (possibly UTD Rooms too?)

## Tasks
- [x] UI
- Filters
  - [x] Desktop side panel
  - [x] Filter bar
  - [x] URL search params handling
  - [ ] Search bar query
  - [ ] Sort and pagination
  - [ ] Button to clear all filters
- Event cards
  - [ ] Responsive design
  - [ ] List view (in addition to current gallery view)
- [ ] Update `findByFilters` event API procedure (currently, event search results are disabled)
- [ ] Refactor
- Future PRs: calendar tab, subscribing/syncing events calendar to user's personal calendar app (possibly move this to `/community` so it's only for events relevant to user)

### Bugs
- [ ] The filter bar has a pre-determined order, i.e. adding new filters does not add them to the end (which is the desired behavior). To fix this, rework how the `selectedFilters` is calculated (probably turn it into a store or a state) and utilize `FilterSearchParams`'s callback functions for `onSet()`, `onDelete()`, etc.
- [ ] When navigating to the `/events` page, the tab at the top of the page doesn't get set correctly.
- [ ] The exclude button in FilterList should always be visible on touch screens, rather than only on hover
- [ ] I added hotkeys for toggling the sidebar (`\`) and changing pages (arrow keys). Currently, these are only active when the current target is body, which is quite strict. Modify this so it's only active when an input field is not selected

Anything else that seems to be broken is not implemented

## Preview

Note the search results are nonfunctional

URL route for following example: `/events?hideRegistered&tags=Academic+Interests%2CEducational&dateStart=2026-03-09&dateEnd=2026-03-13&location=on-campus%2Conline&location%21=off-campus`

image


Isoscelestial:
Moved to #651 (somehow this PR's source branch is on my fork? I genuinely don't know how that happened)

PR #656, opened 2026-03-18
use clubRecord instead of club when checking calendarId for `disableSync`

AbhiramTadepalli:
We're in the events update, so referencing club causes it to error. clubRecord exists already and has the calendarId as a string.


TyHil (approved):
LGTM IG

PR #661, opened 2026-03-29
fix: allow editing past GCal events after unsync

### LINKED ISSUE #643: Allow editing past GCal events after unsync

GCal events are uneditable on Clubs. However, when I unsync my GCal, past events are kept for posterity, but I should then be able to edit/delete those events.

### What happened?

Past events after unsyncing gcal are uneditable and undeletable.

### Change to make

In the `disableSync` procedure in `src\server\api\routers\event.ts`, if `input.keepPastEvents` is true then set the `google` param of all past events with `google===true` to `false`. This will make those events editable.

Verify that when resyncing the same calendar, these events are "reclaimed" by the google calendar sync, as in they once again have their `google` param set to `true`.

### Version

clubs.utdnebula.com

_No response_

### Relevant log output

```shell

```

### Code of Conduct

SinhSinhAn:
## Summary
Fixes #643

When you unsync your Google Calendar but keep past events, those events were still locked as "Google events," so you couldn't edit or delete them. The whole point of keeping them is so you can manage them yourself.

**What this PR does:**
- After unsyncing, past events that are kept now have their `google` flag set to `false`, making them fully editable
- If you resync the same calendar later, those events get picked back up by Google Calendar sync automatically

### DIFF
--- a/src/server/api/routers/event.ts
+++ b/src/server/api/routers/event.ts
@@ -604,6 +604,26 @@ export const eventRouter = createTRPCRouter({
           ),
         );
 
+      // Mark kept past events as non-google so they become editable
+      if (input.keepPastEvents) {
+        await ctx.db
+          .update(events)
+          .set({ google: false })
+          .where(
+            and(
+              eq(events.clubId, input.clubId),
+              eq(events.google, true),
+              lte(events.startTime, new Date()),
+              clubRecord.calendarId
+                ? or(
+                    eq(events.calendarId, clubRecord.calendarId),
+                    isNull(events.calendarId),
+                  )
+                : undefined,
+            ),
+          );
+      }
+
       // remove google calendar info from the club
       await ctx.db
         .update(club)
--- a/src/utils/calendar.ts
+++ b/src/utils/calendar.ts
@@ -159,6 +159,7 @@ export async function syncCalendar(
                   'createdAt',
                   'updatedAt',
                   'calendarId',
+                  'google',
                 ]),
               });
           }


TyHil (approved):
This looks good. I'm gonna request @AbhiramTadepalli for another review since he did a lot of the work on calendar sync.

SinhSinhAn:
@AbhiramTadepalli pookie bear I need that review ^-^

AbhiramTadepalli (changes requested):
Can you check if resyncing a calendar after un-syncing changes the events to `{google: true}` again?

AbhiramTadepalli:
Also might not be in the scope of the issue, but do we prohibit edits to gcal events? Or is it a frontend-only check?

SinhSinhAn:
> Can you check if resyncing a calendar after un-syncing changes the events to `{google: true}` again?

If it's a full resync, GenerateEvent() make sure that google:true stays true, The event would stay locked, but htis is super rare so I highly doubt this would happen 

Our everyday Case is the incremental Sync which allows google to stay false and stay editable, so to answer your question Yes, if you resync the same calendar and Google returns those events (on a full reset sync), they will be set back to google: true

SinhSinhAn:
> Also might not be in the scope of the issue, but do we prohibit edits to gcal events? Or is it a frontend-only check?

t's frontend-only. The backend event.update mutation only checks isUserOfficer so there's no server-side guard on the google flag. So theoretically if someone bypassed the UI they could edit a live Google Calendar event and the DB would accept it without complaint. So I guess we can make a follow up issue for a proper fix that would add 
'and(eq(events.id, id)', eq(events.google, false)) to the .where() clause, or throw if the fetched event has google: true.

I'll go ahead and make this an issue so some other person can pick up it if you're okay with it

AbhiramTadepalli (approved):
go ahead and make the issue, this lgtm!

SinhSinhAn:
@BK2004 review my pr pookie bear (,,>﹏<,,)👉👈

PR #651, opened 2026-03-10
Redesigned events directory

Isoscelestial:
Implements redesigned events directory design from [Figma](https://www.figma.com/design/snmEoRUgfjAZYSc6hqySkY/Wireframes-for-Jupiter?node-id=7541-7055&t=X3mhMQM4SusRiU5i-11)

## Overview
- Front-end
  - Completely redesigned `/events` page
  - Created lots of components related to filtering, which will be useful for UTD Trends
  - Adds collapsible panels from #630
  - Add "list" variant to `EventCard` and tweaked to be more responsive
- TRPC
  - Completely reworked `findByFilter` procedure in event router
  - Renamed `countByClubId` procedure in event router to `count`, and updated to accept `clubId` optionally
- Database
  - Add SQL index on events table (courtesy of @nl32)
- Utilities
  - `src/utils/eventFilter.ts`: Updated to match redesigned `/events` page
  - `src/utils/FilterSearchParams.ts`: Wrapper class for URLSearchParams to add type safety and side effect callback functions
  - `src/utils/floatingActionButtons.ts`: Unused utility that can hide injected floating action buttons. I kept it in case someone might find it useful, but it's technically unused so...
  - `src/utils/searchParams.ts`: Utility functions and classes for safely setting URL search parameters
  - `src/utils/useStable.ts`: Utility function to stabilize a variable; i.e. ensure a variable's reference doesn't change if an object's or array's values are deeply equivalent. This is computationally expensive and might be redundant in some places now, so I may attempt to remove this later.

## Tasks
- [x] UI
- Filters
  - [x] Desktop side panel
  - [x] Filter bar
  - [x] URL search params handling
  - [x] Search bar query
  - [x] Sort and pagination
  - [x] Button to clear all filters
- Event cards
  - [x] Responsive design
  - [x] List view (in addition to current gallery view)
- [x] Update `findByFilters` event API procedure
- [x] Refactor
- Future PRs:
  - calendar tab
  - subscribing/syncing events calendar to user's personal calendar app
  - add "modality" column to `events` table, which is typed as the `eventLocationFilterEnum` enum in `eventFilter.ts`

### Bugs
- [ ] The filter bar has a pre-determined order, i.e. adding new filters does not add them to the end (which is the desired behavior). To fix this, rework how the `selectedFilters` is calculated (probably turn it into a store or a state) and utilize `FilterSearchParams`'s callback functions for `onSet()`, `onDelete()`, etc.
[...]

PR #675, opened 2026-04-06
feat: store BaseHeader height in a CSS variable

### LINKED ISSUE #671: Store BaseHeader height in a CSS variable
## Overview

Some CSS properties require an offset that is equivalent to the height of the header. Normally, there should be a CSS variable corresponding to the header's height, but we don't have that yet. Add this. Then, add this variable to places where it should be used (see https://github.com/UTDNebula/utd-clubs/pull/651#discussion_r3037803954)

## Proposal

Use a ResizeObserver in `BaseHeader.tsx` to measure the height of the base header component, then do something like this:
```
document.documentElement.style.setProperty('--base-header-height', `${height}px`)`
```

Alternatively, if we assume that BaseHeader's height will never change (although it does on extremely small screens), we could just give the variable a value of 68px (which corresponds to 17 in Tailwind btw)

SupaSeeka:
Closes #671 

Added the ResizeObserver in BaseHeader.tsx.


src/components/header/BaseHeader.tsx
```diff
@@ -135,9 +142,32 @@ export const BaseHeader = ({
...
 
+  const headerRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    const header = headerRef.current;
+    if (!header) return;
+
+    const observer = new ResizeObserver(([entry]) => {
+      const height =
+        entry.borderBoxSize[0]?.blockSize ?? entry.contentRect.height;
```
Isoscelestial: TypeScript complains about `entry` possibly being undefined, add optional chaining to `entry`
```suggestion
        entry?.borderBoxSize[0]?.blockSize ?? entry?.contentRect.height;
```

src/components/header/BaseHeader.tsx
```diff
@@ -135,9 +142,32 @@ export const BaseHeader = ({
 
   const [openCompactSearchBar, setOpenCompactSearchBar] = useState(false);
 
+  const headerRef = useRef<HTMLDivElement>(null);
+
```
Isoscelestial, line 214: Add comment here to explain what this does

```suggestion
  // Dynamically stores header height into a CSS variable called `--base-header-height`.
  // Useful for elements that need top offsets with `position: fixed` or `scroll-margin-top`.
```

Isoscelestial (changes requested):
I wasn't expecting the issue to be done so fast lol, nice

I forgot to put this in the original issue, but I want to merge the [most recent version of BaseHeader from UTD Trends](https://github.com/UTDNebula/utd-trends/blob/develop/src/components/navigation/Header/BaseHeader.tsx) into this file. Would you like to do this, or would you be comfortable with me adding a commit to this PR?

Also this PR is waiting on #651

src/components/header/BaseHeader.tsx
SupaSeeka: Yep, these look good.

SupaSeeka:
Happy for you to add that commit to the PR.

Isoscelestial:
@JameKrame showed interest in taking this issue up, so I'm assigning it to him.

What to do:
- Copy/paste the [BaseHeader from Trends](https://github.com/UTDNebula/utd-trends/blob/develop/src/components/navigation/Header/BaseHeader.tsx) into `BaseHeader.tsx` in this PR, making sure to keep not overwrite the changes that SupaSeeka added.
- Modify `Header.tsx` to use the new props (you can view the one on [Trends](https://github.com/UTDNebula/utd-trends/blob/develop/src/components/navigation/Header/Header.tsx) to see what it should sorta look like)

AbhiramTadepalli:
@Isoscelestial should this go in the Library now instead?

Isoscelestial:
@JameKrame Are you available to work on this sometime this week? If not, absolutely no worry cause I'm more than happy to finish this PR! We just need this in soon.

Isoscelestial (approved):
Thanks @SupaSeeka for adding this!

Sorry @JameKrame for me taking over. It's just we're doing a lot of refactoring this upcoming week, and this PR is a bit of a blocker. Don't worry, next semester we have tons of other issues to work on lol

PR #680, opened 2026-04-09
feature: show tag count on homepage

### LINKED ISSUE #601: Show tag counts on the homepage
Show the number of clubs that are using a tag on the homepage. We could display it like "Tag (13)" or use a [MUI Badge](https://mui.com/material-ui/react-badge/)

Maybe sort the list by this?

Yosias1706:
Updated the tag system to show the number of clubs using a tag and sort the tags list by this on the homepage and the manage club page. Alphabetical order is the tiebreaker for tags with the same number of clubs using them.
**Homepage example:**
Screenshot 2026-04-08 at 8 48 09 PM

**Club Manage Page example:**
Screenshot 2026-04-08 at 8 50 03 PM
I updated club.ts to return the whole tag object including the count instead of just the tag name, and updated the other files to handle this new data including a mapping in ClubTagEdit to stop the Autocomplete search functionality from breaking due to expecting a string instead of an object. renderOption is used to display the tag counts in the dropdown on the manage clubs page. I didn't know if I was supposed to put the count inside a tag when its selected in the search bar as well so I didn't.

### DIFF
--- a/src/components/AllTags.tsx
+++ b/src/components/AllTags.tsx
@@ -71,7 +71,16 @@ export const AllTags = ({
       >
         {options.map((option) => (
           <MenuItem key={option.tag} onClick={() => handleChange(option.tag)}>
-            <ListItemText primary={`${option.tag} (${option.count})`} />
+            <ListItemText
+              primary={
+                <>
+                  <span>{option.tag}</span>
+                  <span className="ml-2 text-sm text-slate-600 dark:text-slate-400">
+                    ({option.count})
+                  </span>
+                </>
+              }
+            />
           </MenuItem>
         ))}
       </Menu>
--- a/src/components/AllTags.tsx
+++ b/src/components/AllTags.tsx
@@ -11,7 +11,7 @@ export const AllTags = ({
   options = [],
   label = '•••',
 }: {
-  options?: string[];
+  options?: { tag: string; count: number }[];
   label?: string;
 }) => {
   const { tags, setTags, setShouldFocus } = useSearchStore((s) => s);
@@ -48,8 +48,6 @@ export const AllTags = ({
     setShouldFocus(true);
   };
 
-  const sorted = options.sort((a, b) => a.localeCompare(b));
-
   return (
     <>
       <Button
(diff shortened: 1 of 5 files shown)


src/components/TagPill.tsx
```diff
@@ -22,9 +24,11 @@ export const TagPill = ({
     });
   }
 
+  const displayName = count !== undefined ? `${name} (${count})` : name;
+
   return (
     <TagChip
-      tag={name}
+      tag={displayName}
```
Isoscelestial: In yesterday's meeting, we felt that showing the count on the chips added additional clutter. Could you try moving the count to a tooltip instead?

src/components/manage/form/ClubTagEdit.tsx
```diff
@@ -41,6 +45,18 @@ export const ClubTagEdit = ({
...
       options={allTags ?? []}
+      renderOption={(props, option) => {
+        const { key, ...otherProps } = props;
+        const original = rawTags?.find(
+          (t: { tag: string; count: number }) => t.tag === option,
+        );
+
+        return (
+          <li key={key} {...otherProps}>
+            {option} {original ? `(${original.count})` : ''}
```
Isoscelestial: Could you make the count use a similar design to the Min Letter Grade filter [on UTD Trends](https://trends.utdnebula.com/dashboard?searchTerms=CS+1200)? Here's where those lines of code are: https://github.com/UTDNebula/utd-trends/blob/9754fe035c760bb2dd9973930602feac5439a5a9/src/components/search/Filters/Filters.tsx#L294-L297

src/components/AllTags.tsx
```diff
@@ -71,9 +69,9 @@ export const AllTags = ({
           list: { onClick: (e: React.MouseEvent) => e.stopPropagation() },
         }}
       >
-        {sorted.map((tag) => (
-          <MenuItem key={tag} onClick={() => handleChange(tag)}>
-            <ListItemText primary={tag} />
+        {options.map((option) => (
+          <MenuItem key={option.tag} onClick={() => handleChange(option.tag)}>
+            <ListItemText primary={`${option.tag} (${option.count})`} />
```
Isoscelestial: Also do that here

Isoscelestial (changes requested):
Nice work! Just a few design tweaks

Isoscelestial (approved):
Looks good to me, thanks!

BK2004 (approved):

PR #685, opened 2026-04-21
Changed "There are no upcoming events" to "This club has not posted info about upcoming events"

### LINKED ISSUE #602: Change "There are no upcoming events"
# Proposal

When viewing a club that has not configured events such as [ACM](https://clubs.utdnebula.com/directory/association-for-computing-machinery), the text under Upcoming Events says "There are no upcoming events". 

Image

For clubs that are missing from UTD Clubs such as the [ASL organization](https://clubs.utdnebula.com/directory/american-sign-language-organization), it instead says "No info about upcoming events"

Image

I understand if someone disagrees and I'd love to hear feedback, but I think "No info about upcoming events" should be used for both. "There are no upcoming events" can make the club sound abandoned instead of just not setting up their events on UTD clubs.

T0MN0M:


src/components/club/listing/ClubBody.tsx
```diff
@@ -64,7 +64,7 @@ const ClubBody = async ({
           emptyText={
             club.updatedAt == null || club.updatedAt < oneYearAgo
               ? 'No info about upcoming events'
-              : 'There are no upcoming events'
+              : 'This club has not posted info about upcoming events'
```
Isoscelestial: Use a contraction to reduce word count and make it easier to read

```suggestion
              : "This club hasn't posted info about upcoming events"
```

Isoscelestial (changes requested):
I personally like the verbosity of the words you used (lol I'm a stickler for precision). Think it'd be good to have another opinion on this, though.

src/components/club/listing/ClubBody.tsx
justinschwerdtfeger: I also slightly prefer the version with the contraction, and I really like the precision.

T0MN0M: True, the contraction will make the line easier to read.

Isoscelestial (approved):
Thanks for the change!

justinschwerdtfeger (approved):
Looks Great! Good job! 🙌 🥳

PR #686, opened 2026-04-22
feat: composite PK on events table (fixes #611)

### LINKED ISSUE #611: Make events use a composite key
# Overview

Spinning off of #541 & #574, events definitely need to be defined with a primary key of `{event_id, club_id}`

# Why?

With the new Gcal integration changes in #541, a calendar's event can potentially be used by 2 clubs.
> * One edge case is when Club A desyncs a calendar but keeps past events and club B syncs, past events are now gone from club A (club B "owns" them now)

## Supplementary Information
Ask me and I can provide more info if you need

SinhSinhAn:
## What's the problem?

Right now the `events` table uses just `id` as its primary key. When two clubs link the **same** Google Calendar, the calendar sync tries to insert the same event ID for both clubs and one of them loses. Whichever club syncs second either overwrites the first club's event or gets a conflict error.

## What does this PR do?

Changes the events primary key from `(id)` to `(id, club_id)`, so the same Google Calendar event can exist under multiple clubs without stepping on each other.

### Schema

- **`events`** table: primary key is now `(id, club_id)` instead of just `(id)`
- **`user_metadata_to_events`** table: added a `club_id` column, primary key is now `(user_id, event_id, club_id)`, and the foreign key points to `events(id, club_id)` instead of just `events(id)`

### Backend (tRPC routers + calendar sync)

- **Event router**: `joinedEvent`, `registerState`, `toggleRegistration`, `update`, and `delete` all now require and filter by `clubId`
- **Admin router**: `deleteEvent` now requires `clubId` and scopes both the registration cleanup and the event delete by `(id, club_id)`
- **Calendar sync** (`calendar.ts`): upsert conflict target changed to `[id, clubId]`, and soft-deletes of cancelled events are now scoped by `clubId` so one club's sync can't accidentally mark another club's copy as deleted

### Frontend

- **EventRegisterButton**: passes `clubId` to `registerState`, `toggleRegistration`, and all cache key references
- **EventDeleteButton**: passes `clubId` alongside `id` when calling delete

### Migration

The migration (`20260419000000_events_composite_pk.sql`) does this in order:
1. Drops the old FK from `user_metadata_to_events` to `events`
2. Swaps the events PK from `(id)` to `(id, club_id)`
3. Adds `club_id` to `user_metadata_to_events` (nullable at first)
4. Backfills `club_id` from the `events` table
5. Deletes any orphaned registrations that don't match an event
6. Makes `club_id` NOT NULL
7. Sets the new composite PK and FK

Closes #611

### DIFF
--- a/src/components/events/EventDeleteButton.tsx
+++ b/src/components/events/EventDeleteButton.tsx
@@ -81,7 +81,7 @@ export default function EventDeleteButton({
           </>
         }
         onConfirm={() => {
-          deleteEvent.mutate({ id: event.id });
+          deleteEvent.mutate({ id: event.id, clubId: event.clubId });
         }}
         loading={deleteEvent.isPending}
       />
--- a/src/components/events/EventRegisterButton.tsx
+++ b/src/components/events/EventRegisterButton.tsx
@@ -32,15 +32,15 @@ const EventRegisterButton = ({
   const api = useTRPC();
   const queryClient = useQueryClient();
   const { data: registerState, isPending } = useQuery(
-    api.event.registerState.queryOptions({ id: eventId }),
+    api.event.registerState.queryOptions({ id: eventId, clubId }),
   );
 
   const toggleRegistration = useMutation(
     api.event.toggleRegistration.mutationOptions({
-      onMutate: async ({ id }) => {
+      onMutate: async ({ id, clubId }) => {
         const queryKey = [
           ['event', 'registerState'],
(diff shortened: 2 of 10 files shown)


Isoscelestial:
We'd like to discuss alternative solutions to resolving the Google Calendar re-sync edgecase that don't involve using a composite key, so this PR is on hold for now. Thanks for working on it though

PR #689, opened 2026-04-28
Extracted the time zone from the system onto the event times

### LINKED ISSUE #608: Specify timezones or convert to user's timezone
# Overview

All the events are generally given in central time.

# Proposal

Either
(a) allow club managers to specify timezone for the event start/end time + convert to end-user's time zone
(b) specify to manger & end-user that all times are in Central Time.

## Supplementary Information
In the DB, everything is stored in UTC

SGcoder120:
Adjusted timezone of events based on system settings.

### DIFF
--- a/src/components/events/ClientEventTime.tsx
+++ b/src/components/events/ClientEventTime.tsx
@@ -17,12 +17,14 @@ const ClientEventTime = ({ startTime, endTime }: ClientEventTimeProps) => {
     return null;
   }
 
-  const tzStartTime = new TZDateMini(startTime, 'America/Chicago');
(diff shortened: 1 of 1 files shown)


Isoscelestial:
Can confirm this works, though we'd need to add it to the event page and a few other places I believe.

However, we decided in our meeting a couple weeks ago ([meeting notes](https://nebula-labs.atlassian.net/wiki/spaces/UJ/pages/1096581121/2026-04-14+Meeting+notes)) that we'd keep times in Central Time because that's the local time of the event (i.e. most events on Clubs happen at UTD, which is in Central Time anyway). Gonna put this PR on hold for now, but we probably will pursue a different direction.

PR #690, opened 2026-05-03
fix: enforce google flag on event update mutation

### LINKED ISSUE #683: Backend does not enforce google flag on event edits
## Summary

The `event.update` backend mutation does not enforce the `google` flag on events. This means the guard that prevents editing Google Calendar synced events only exists on the frontend. Anyone who bypasses the UI can edit a GCal event directly through the API, and the server will accept it.

---

## Background: What is the `google` flag?

When a club syncs their Google Calendar, events are pulled in and stored in the database with `google: true`. This flag is meant to signal that the event is owned by Google Calendar and should not be edited manually, since any changes would be overwritten the next time the calendar syncs.

The frontend respects this by hiding or disabling the edit button for events where `google: true`. However, this check does not exist on the backend.

---

## The Problem

In `src/server/api/routers/event.ts`, the `event.update` mutation looks like this (simplified):

```ts
update: protectedProcedure
  .input(editEventSchema)
  .mutation(async ({ input, ctx }) => {
    const isOfficer = await isUserOfficer(ctx.session.user.id, input.clubId);
    if (!isOfficer) {
      throw new TRPCError({ code: 'UNAUTHORIZED', ... });
    }

    await ctx.db
      .update(events)
      .set({ name: data.name, location: data.location, ... })
      .where(eq(events.id, id));
  }),
```

The only check here is whether the user is an officer. There is no check for whether the event has `google: true`. This means:

- A club officer could construct a direct API request (bypassing the frontend entirely)
- The server would happily update a live Google Calendar event in the database
- The next calendar sync would overwrite those changes anyway, or the record would silently diverge from Google Calendar

This is a security/data integrity gap. The frontend alone should never be the only thing standing between a user and a protected write.

---

## How to Fix

There are two approaches. Pick one.

### Option A: Guard in the `.where()` clause (recommended, fewer queries)

Add `eq(events.google, false)` to the `.where()` so the update only runs if the event is not GCal-owned:

```ts
const res = await ctx.db
  .update(events)
  .set({ ... })
  .where(and(eq(events.id, id), eq(events.google, false)))
  .returning({ id: events.id });

if (res.length === 0) {
  throw new TRPCError({
    code: 'FORBIDDEN',
    message: 'Cannot edit a Google Calendar event directly.',
  });
}
```

If the event has `google: true`, the update will match zero rows and you throw before returning.

### Option B: Fetch first, then throw (more explicit)

Fetch the event before attempting the update and check the flag:

```ts
const existing = await ctx.db.query.events.findFirst({
  where: eq(events.id, id),
});

if (existing?.google) {
  throw new TRPCError({
    code: 'FORBIDDEN',
    message: 'Cannot edit a Google Calendar event directly.',
  });
}

await ctx.db.update(events).set({ ... }).where(eq(events.id, id));
```

This is slightly more verbose but easier to read and easier to extend with additional checks later.

---

## Files to Look At

- `src/server/api/routers/event.ts` -- the `update` mutation is where the fix goes
- `src/server/db/schema/events.ts` -- shows the `google` boolean column on the events table

---

## Context

This was surfaced during a code review on PR #661 by @AbhiramTadepalli. It is out of scope for that PR but worth fixing as a follow-up so that the backend enforces the same rules the frontend already shows to users.

SinhSinhAn:
## Summary
Closes #683

- Adds `eq(events.google, false)` to the `.where()` clause in the `event.update` mutation so Google Calendar synced events cannot be modified through the API
- Previously only the frontend hid the edit button for GCal events; the backend accepted edits regardless

### DIFF
--- a/src/server/api/routers/event.ts
+++ b/src/server/api/routers/event.ts
@@ -462,12 +462,30 @@ export const eventRouter = createTRPCRouter({
         .where(and(eq(events.id, id), eq(events.google, false)))
         .returning({ id: events.id });
 
-      if (res.length === 0)
-        throw new TRPCError({
-          code: 'FORBIDDEN',
-          message:
-            'Cannot edit a Google Calendar event directly, or event not found.',
+      if (res.length === 0) {
+        const existing = await ctx.db.query.events.findFirst({
+          where: (e) => eq(e.id, id),
         });
+
+        if (!existing) {
+          throw new TRPCError({
+            code: 'NOT_FOUND',
+            message: 'Event not found.',
+          });
+        }
+
+        if (existing.google) {
+          throw new TRPCError({
+            code: 'FORBIDDEN',
+            message: 'Cannot edit a Google Calendar event directly.',
+          });
+        }
+
+        throw new TRPCError({
+          code: 'INTERNAL_SERVER_ERROR',
+          message: 'Failed to update event.',
+        });
+      }
       return res[0]?.id;
     }),
   delete: protectedProcedure
--- a/src/server/api/routers/event.ts
+++ b/src/server/api/routers/event.ts
@@ -459,13 +459,14 @@ export const eventRouter = createTRPCRouter({
           image: data.image,
           updatedAt: new Date(),
         })
-        .where(eq(events.id, id))
+        .where(and(eq(events.id, id), eq(events.google, false)))
         .returning({ id: events.id });
 
-      if (res.length == 0)
+      if (res.length === 0)
         throw new TRPCError({
-          code: 'INTERNAL_SERVER_ERROR',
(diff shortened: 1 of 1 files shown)


Isoscelestial (changes requested):
Instead of flat out replacing the `INTERNAL_SERVER_ERROR` error with a `FORBIDDEN` error, could you implement logic such that whenever `res.length === 0` (i.e. there is an error), it will query the database for the existing item and will return one of the following errors:
- `NOT_FOUND` if doesn't exist
- `FORBIDDEN` if google is false
- `INTERNAL_SERVER_ERROR` for fallback error that you removed

Isoscelestial (approved):

PR #696, opened 2026-05-09
Only show tags from approved clubs; Admin tags page redesign

### LINKED ISSUE #631: Pull tag list only from approved clubs

The list of all tags does not include tags from pending or rejected clubs

### What happened?

It does do that

### Version

clubs.utdnebula.com

_No response_

### Relevant log output

```shell

```

### Code of Conduct

Isoscelestial:
Please ignore how I accidentally pushed these changes to the develop branch at first... sorry

- Resolves #631 
- Moves `changeTags` procedure to the admin router, which is where it belongs
- Add `refreshTags` procedure in admin router
- Also redesigns admin tags page cause... I'm not sure why I did that, honestly.

redesigned tags list


TyHil (dismissed):
Looks like this is also making the change to only show tags from approved clubs. Would also be nice to regenerate the tags list when a club is deleted. Could make an issue or do in this PR but just an FYI

Isoscelestial:
> Could make an issue or do in this PR but just an FYI

That is a good idea. This would be a good first issue IMO, I will make it an issue for the recruits next semester

PR #698, opened 2026-05-23
Update packages for Summer 2026

Isoscelestial:
## Changes
- update to Typescript 6
  - remove deprecated stuff from `tsconfig.json`
- update most packages to their latest versions
- remove some unused packages:
  - `formidable`
  - `react-hook-form`
- remove useless `.mocharc` file
- update `tests/tests.ts` to finally get rid of those misc. type errors we've had for a while lol
- disable the ESLint `@next/next/no-img-element` rule in `opengraph-image` files
- add `type:check` and `drizzle:studio` npm scripts
- update `lint-check` GitHub workflow to also validate types

## Notes
- We cannot update ESLint to version 10.0 because [`eslint-plugin-jsx-a11y` does not support it](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/1075). However, that package hasn't been updated in 1.5 years, so we may need to consider either removing it or finding a replacement.
  - Once we update ESLint to version 10.0, we should remove the `@eslint/eslintrc` package since we don't use a `.eslintrc` file
- I didn't update `@google-analytics/data` and `@google/genai` because I don't want to hassle with breaking changes
- I didn't update MUI yet because I am fearful there could be a lot of breaking changes that will require a *lot* of testing, and I'd rather do that in a separate PR. Plus, it's worth waiting until MUI 9 matures a little, I guess?
- [ts-node](https://www.npmjs.com/package/ts-node) hasn't been updated in several years, do we need to switch to [tsx](https://www.npmjs.com/package/tsx)?


TyHil:
Is `npm run type:check` something to be added to .github/workflows/lint-check.yml? Or would it just always fail when `npm run build` fails (and thus be covered by the Vercel deployment)?

Isoscelestial:
> Is `npm run type:check` something to be added to .github/workflows/lint-check.yml? Or would it just always fail when `npm run build` fails (and thus be covered by the Vercel deployment)?

We should add the script to the workflow, since we'll need it to check type safety in `/tests/` and `/scripts/` once we start adding tests with Jest. Good catch, I'll add it.

TyHil:
Is this ready to merge?

Isoscelestial:
Yup, ready to merge! I was going to try updating MUI, but as I feared there's a ton of breaking changes so that can wait till later this summer.

TyHil (approved):
Make sure to remove the "Run ESLint + Prettier" check and add your "Run ESLint + Prettier + TypeScript" check in both the `develop` and `main` branch protection rules under "Require status checks to pass before merging" at https://github.com/UTDNebula/utd-clubs/settings/branches after merge :)

Isoscelestial:
Done! Thanks for letting me know I have to do that

PR #706, opened 2026-06-22
Refactored snackbar system

Isoscelestial:
I was gonna try again to add stacked snackbars, but I changed my mind and ended up just doing a bunch of refactoring. Ah, well. If we do wanna add that in the future, at least the snackbar code isn't a mess anymore.

Not a lot of functional changes here, mostly just making the system more robust.

### Snackbar system
- refactor snackbar system to split bloated `/components/global/Snackbar.tsx` file to multiple files in `/utils/snackbar/`
- add `setSnackbarWithPreset()` and `closeSnackbar()` functions to `SnackbarContext` and globally exported functions
- change `closeOn` option for Snackbars to use an object of booleans instead of an array, so that it can be merged with default values
- add `id` option for Snackbars, so that `closeSnackbar()` can filter for that (to prevent you from programmatically closing a snackbar you didn't mean to)
- tweak `SnackbarPresets` to have more sensible options

### Events Directory
- add `SSRSnackbarWrapper` to snackbar system, which allows SSR components to display a snackbar on the client. This is useful when there is a TRPC procedure error on the server and we want to display an error to the user, but don't want to SSR a banner into the page. I implemented this idea on the events directory.
  - To test the generic snackbar, throw a `TRPCError` on line 280 of `/server/api/routers/event.ts`
- inform user through snackbar when clearing event directory filters fixes server-side error
  - To test the filter warning snackbar, throw a `TRPCError` on line 457 of `/server/api/routers/event.ts`, then select "Your Clubs" in the filters
- For the "this filter only works when signed in" filter, I made it close whenever reselecting the default filter option that doesn't require being signed in.

All other changes are updates to imports and renamed snackbar presets.

## Discussion

Should I move the snackbar system to `/modules/` instead of `/utils/`? I don't even know what the difference is between the two lol

## To do
- Copy this over to https://github.com/UTDNebula/utd-notebook (that's why I requested @ZeyadAbbas to review)
- Do similar refactoring for RegisterModal, which is arguably even worse because that was one of the first features I ever contributed :sob:


TyHil (dismissed):
Not a super thorough review but looks good.

`modules` is p much just what `utils` is called on Trends and someone working on both may have just accidentally copied it over.

AbhiramTadepalli (dismissed):
love it! Tried to test it on different pages & events page works well too!

PR #709, opened 2026-06-24
Refactored register modal

Isoscelestial:
Similar to #706 

### File organization
- Rename "register modal" to "login modal" across codebase (to reduce confusion with event registration features)
- Move appropriate files to a `utils/loginModal/` folder:
  - `/components/global/RegisterModal`
  - `/components/global/RegisterModalProvider`
  - `/app/auth/ProviderButtons`
  - `/icons/AuthIcons`
- Split login modal to multiple files

### Code
- Replaced `setShowRegisterModal()` with `openLoginModal()` and `closeLoginModal()`
- Ability to pass a callbackURL to `openLoginModal()`
- `onClose` if user closes login modal after it's opened. Not used anywhere, just might be useful idk
- Added globally exported `openLoginModal()` and `closeLoginModal()` functions if you don't wanna use the `useLoginModal` hook

### Design
- Use "Continue with" language instead of "Sign in with" since the buttons function as both login and registration
- Fix icon positions in button
- Reduce usage of slate colors and Bai Jamburee font
- Make it clearer where faculty/staff are supposed to click
- Add outlines to buttons for contrast and cause ngl it looks cool

**Before:**
login modal before

**After:**
login modal after

## To do
- Probably move to another folder, assuming we do @TyHil's idea of a Nebula Labs component repository
- Copy over to https://github.com/UTDNebula/utd-notebook


AbhiramTadepalli (dismissed):
seems to work for me :)

src/utils/loginModal/global.ts
```diff
@@ -0,0 +1,49 @@
...
+    globalLoginModalFunctions.openLoginModal = functions.openLoginModal;
+    globalLoginModalFunctions.closeLoginModal = functions.closeLoginModal;
+
+    return () => {
+      globalLoginModalFunctions.openLoginModal =
+        globalLoginModalFunctionsUnmounted;
+      globalLoginModalFunctions.closeLoginModal =
+        globalLoginModalFunctionsUnmounted;
+    };
+  });
```
AbhiramTadepalli: should there be a dependency array?

Isoscelestial: Yes, whoops. Good catch

PR #713, opened 2026-07-21
Update NEBULA_API_STORAGE_BUCKET in README

TyHil:
The bucket is still called jupiter so I shouldn't have changed this in #344


Isoscelestial (approved):
Lol

PR #722, opened 2026-08-16
install `rehype-raw` to parse HTML in MarkdownText.tsx

### LINKED ISSUE #721: Fix formatting of event descriptions imported from GCal

Events from Google Calendar are imported with HTML descriptions. Right now `MarkdownText.tsx` sanitizes it and displays events so that they don't have a line break when needed:
Image

### What happened?

Because the description is stored as:
```
Come work on API &amp; Platform! Open to anyone!<br><br>The Nebula API serves as the authoritative data source for UTD information such as courses, student organizations, and more.<br><br><a href="https://www.google.com/url?q=https://www.utdnebula.com/projects/api&amp;sa=D&amp;source=calendar&amp;usd=2&amp;usg=AOvVaw0ieAq6u-cvguXovq7iRGyz" target="_blank">https://www.utdnebula.com/projects/api</a><br>
```

line breaks are currently ignored, and the subsequent paragraph is merged into the preceding one.

Fix this by reading those line breaks (and other html formatting)

### Version

clubs.utdnebula.com

_No response_

### Relevant log output

```shell

```

### Code of Conduct

AbhiramTadepalli:
Fixed #721 by installing and importing `rehype-raw` before sanitization in `MarkdownText.tsx`

**I need help testing** if this breaks anything for other event descriptions + club descriptions since `ExpandableMarkdownText.tsx` is used in both places

### DIFF
--- a/src/components/MarkdownText.tsx
+++ b/src/components/MarkdownText.tsx
@@ -21,9 +21,12 @@ const MarkdownText = ({
       <Markdown
         remarkPlugins={[remarkGfm]}
         rehypePlugins={[
-          [rehypeExternalLinks, { rel: ['noreferrer'], target: ['_blank'] }],
           rehypeRaw,
           rehypeSanitize,
+          [
+            rehypeExternalLinks,
+            { rel: ['noopener', 'noreferrer'], target: '_blank' },
+          ],
         ]}
       >
         {text}
--- a/src/components/MarkdownText.tsx
+++ b/src/components/MarkdownText.tsx
@@ -1,5 +1,6 @@
 import Markdown from 'react-markdown';
 import rehypeExternalLinks from 'rehype-external-links';
+import rehypeRaw from 'rehype-raw';
 import rehypeSanitize from 'rehype-sanitize';
 import remarkGfm from 'remark-gfm';
 
@@ -21,6 +22,7 @@ const MarkdownText = ({
         remarkPlugins={[remarkGfm]}
         rehypePlugins={[
           [rehypeExternalLinks, { rel: ['noreferrer'], target: ['_blank'] }],
+          rehypeRaw,
           rehypeSanitize,
         ]}
       >
--- a/package-lock.json
+++ b/package-lock.json
@@ -50,6 +50,7 @@
         "react-dom": "^19.2.6",
         "react-markdown": "^10.1.0",
         "rehype-external-links": "^3.0.0",
+        "rehype-raw": "^7.0.0",
         "rehype-sanitize": "^6.0.0",
         "remark": "^15.0.1",
         "remark-gfm": "^4.0.1",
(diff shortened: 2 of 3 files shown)


justinschwerdtfeger:
I'm working on reviewing this right now

justinschwerdtfeger:
So far in my testing, bold, custom links, italics, ordered and unordered lists work from Google calendar, but underline does not as they are removed by rehypeSanitize. We could make `<u>` not sanitized, but markdown doesn't support underline. I lean towards leaving as it is

 We also don't display google drive attachments, but I don't think we even could

src/components/MarkdownText.tsx
```diff
@@ -21,6 +22,7 @@ const MarkdownText = ({
         remarkPlugins={[remarkGfm]}
         rehypePlugins={[
           [rehypeExternalLinks, { rel: ['noreferrer'], target: ['_blank'] }],
+          rehypeRaw,
           rehypeSanitize,
```
justinschwerdtfeger, line 25: In my testing, the  `target: [_blank]` doesn't work. This is because rehypeExternalLinks is before rehypeSanitize. I think we should change the order

justinschwerdtfeger (changes requested):

src/components/MarkdownText.tsx
justinschwerdtfeger: I can fix this for you if you agree with the change

justinschwerdtfeger:
I couldn't find anything that breaks in my testing. Looks good!

AbhiramTadepalli:
> So far in my testing, bold, custom links, italics, ordered and unordered lists work from Google calendar, but underline does not as they are removed by rehypeSanitize. We could make `<u>` not sanitized, but markdown doesn't support underline. I lean towards leaving as it is


I concur, underlining isn't all too important anyway.


>  We also don't display google drive attachments, but I don't think we even could

The first png/jpg attachment becomes the event's thumbnail.

We should make a follow-up issue for supporting attachments for events (even non-synced ones?)

justinschwerdtfeger:
What about the target: _blank issue?

Isoscelestial:
> We should make a follow-up issue for supporting attachments for events (even non-synced ones?)

I agree, it would be nice to support inline images. We could even support this in club descriptions.

AbhiramTadepalli:
> inline images

idt google calendar allows this, they just allow attachments

src/components/MarkdownText.tsx
AbhiramTadepalli: makes sense, made the change, tested that it opens in a new tab

Isoscelestial (approved):
LGTM!

justinschwerdtfeger (approved):
Looks good to me too!

PR #725, opened 2026-08-20
Update Club Match Form to use wizard

### LINKED ISSUE #720: Update club match to use new form wizard components
Want club match to look more approachable, so let's redesign `/club-match` to use the form wizard design seen in `/get-started` and `/directory/create`

We want to advertise club matching as soon as the semester starts, so would be nice to finish ASAP

justinschwerdtfeger:
### Description

image

image

Resolves #720 
Library changes in this PR
- Add placeholder to FormAutoComplete.tsx
  - The actual component is called FormAutoCompleteFreeSolo. We should probably fix this name difference
- Create FormMultiSelect
- Create FormRadioGroup

Main Changes in this PR
- Update Club Match to use form wizard.
- Rename to ClubMatchForm
- Split ClubMatchForm into steps
- Use snackbar for clubMatch failure

This PR is kinda rushed because I've been busy. I want to make backend changes which I talked about [on discord](https://discord.com/channels/860405177865338890/1347371115504209942/1540126605702205500), but those will be a new issue. I also think the styling could use some work. It looks pretty bland, but definitely better than before. I can't put my finger on what exactly is wrong. The FormMultiSelect's purple on gray looks weird to my eyes, but this same color combo looks alright in other places.

### Testing

- Tested failure (Failure causes snackbar)
- Tested form flow and error states
This PR could use some more testing 

To test, you'll need Google Cloud service account set up for Gemini

### AI Disclosure

Used AI for almost everything component and reviewed everything manually after

### Checklist

- [x] Create this PR
- [x] Perform final self-review


src/systems/clubs/match/steps/InterestsStep.tsx
```diff
@@ -0,0 +1,126 @@
+'use client';
+
+import { withForm } from '@/lib/utils/form';
+import { ClubMatchWizardSchema } from '../clubMatchSchema';
+
+const CATEGORY_OPTIONS = [
```
AbhiramTadepalli, line 7: Can we / should we maintain these dynamically? I know there are a lottt of tags now, though.

AbhiramTadepalli (dismissed):
I didn't review all the form components and steps line-by-line, but it seems to be working great in the browser!

2 nitpicks:
1. Can we make the disclaimer text take up less of the page in mobile?
2. perhaps add the disclaimer text on the results page as well?

Otherwise, all good!

Isoscelestial:
Will review with some UI suggestions soon. I have some thoughts about the form component abstractions, but those can wait since we need this in soon.

> 1. Can we make the disclaimer text take up less of the page in mobile?

The disclaimer is pretty distracting on both desktop and mobile IMO; infact the background color in dark mode has higher emphasis. Would we be fine with either:
1. getting rid of the disclaimer's background and making the text small, or
2. hiding the disclaimer in a modal that appears when you click an Info icon button next to the "Find the perfect club for you" text?

> 2. perhaps add the disclaimer text on the results page as well?

Good idea

src/systems/clubs/match/steps/InterestsStep.tsx
Isoscelestial: The `specificCultures` question is dependent on whether the `Cultural` or `Religious` category is present in the answer. Unsure of best way to handle.

justinschwerdtfeger: I think in general, specific cultures shouldn't be a question (much like gender), and instead, the user should select as many categories as they want from our list of tags. I didn't change this in this PR bc I didn't want to change the functionality of the club match form in this PR.

Isoscelestial (dismissed):
Couldn't find any bugs or regressions, looks good for prod!

justinschwerdtfeger:
What's the force push for?

Isoscelestial:
> What's the force push for?

I pressed GitHub's rebase button, guess it's a feature

justinschwerdtfeger (commented):
I love your work improving the design here! 
Things I noticed: 
- When user is not signed in, selects club match, then signs into an account that has already completed club match, they are sent to `/club-match/form`. I think they should be sent to `/club-match`, though this is a pretty minor problem and can probably be a good first issue.
- This is my fault, but gender identity is a weird question to have on the "involvement" page. 

 I don't think either of these should block merging since its high priority

Isoscelestial (approved):
bazinga

PR #729, opened 2026-08-25
Community page SSR auth error fix

Isoscelestial:
Fix server side auth error when logged out user attempts to access my community page

### DIFF
--- a/src/app/community/page.tsx
+++ b/src/app/community/page.tsx
@@ -1,4 +1,6 @@
 import { type Metadata } from 'next';
+import { headers } from 'next/headers';
+import { auth } from '@/server/auth';
 import ClubEvents from '@/systems/dashboard/ClubEvents';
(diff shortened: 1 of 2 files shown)


src/app/community/page.tsx
```diff
@@ -22,9 +24,14 @@ const Community = async ({
...
   searchParams?: Promise<SearchParams>;
 }) => {
-  const sp = (await searchParams) ?? {};
-  const page = Number(sp.page) || 1;
-  const pageSize = Number(sp.pageSize) || 12;
+  const [session, sp] = await Promise.all([
+    auth.api.getSession({ headers: await headers() }),
+    searchParams,
+  ]);
+  if (!session) return;
```
AbhiramTadepalli, line 31: is it better to `return null`?

AbhiramTadepalli (dismissed):

PR #730, opened 2026-08-25
Add disclaimer about AI hallucination for Club Match

Isoscelestial:
Sometimes, the AI mixes up or hallucinates the club ID in club matching results, which clicking the club sometimes doesn't work or takes you to an incorrect club. Furthermore, the join button for that club would not work properly.

This happens more than we'd like.

What I've tried to do:
- ~~Tell the AI "Double check that the organization ID is valid and corresponds to the correct organization"~~ We'll update the AI prompt in #737 
- Add disclaimer

### DIFF
--- a/src/app/club-match/page.tsx
+++ b/src/app/club-match/page.tsx
@@ -109,7 +109,7 @@ const Page = async () => {
             className="font-display mx-4 my-8 flex items-center justify-center gap-2 text-center text-4xl font-bold"
           >
             Your Top Club Matches
-            <ClubMatchDisclaimer hasSuggestions />
+            <ClubMatchDisclaimer />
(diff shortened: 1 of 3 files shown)


TyHil:
An idea: passing the AI slugs instead of IDs might be easier for it to handle. You could translate back to IDs after the AI reaponds to avoid the case of a club changing their slug and a user coming back to the club match results, but that does seem like a rare occurrence

justinschwerdtfeger (changes requested):
It looks alright, but I think it should maybe be more discoverable. IMO, its not intuitive that the club match page disclaimer isn't the same as before.

justinschwerdtfeger (dismissed):

PR #731, opened 2026-08-29
Memoize admin Filter component to stop excessive re-renders

Isoscelestial:
This page was borderline unusable for months, can't believe the solution was just this :sob:


justinschwerdtfeger:
Just tested, and it seems to fix

AbhiramTadepalli (approved):
makes sense, lgtm, ty!

PR #716, opened 2026-08-09
Restructure codebase

Isoscelestial:
### Description

Tried my best to make commits in a way that preserves Git history, but I'm sure a little bit of git blame history got lost

- Move components and folders to more organized locations, as discussed in [this Discord thread](https://discord.com/channels/860405177865338890/1467979986391666788).
  - Full info: [`docs/project-structure.md`](https://github.com/UTDNebula/utd-clubs/blob/7b8d39044db16bd985211979c19f7d474203a321/docs/project-structure.md)
- Move system-specific utils out of `utils/`
- Split various files, such as `googleCalendar.ts`
- Change `@src/` import prefix to `@/` to match [Trends](https://github.com/UTDNebula/utd-trends) and industry convention (resolves #708)
- Update tRPC routers:
  - TODO: Move stuff in `src/trpc/` folder to better locations. This will be done later cause I need to first study tRPC's documentation to figure out the best way to structure this stuff
  - Shift some procedures to the `user` router
  - Create child routers in the `club`, `event`, and `user` routers
  - Delete the unused `example` router
  - Move input schemas to separate file, and add baseSchema for `id`'s
    - TODO: Create an `officerProcedure` middleware that reads the clubId inputted into a procedure
- Disallow relative paths that move up multiple levels
- Migrate from `ts-node` to `tsx` (resolves #711)
- Add workspace settings and recommended extensions for VS Code (for onboarding new users)
- Begin adding pages to documentation: `nebula-library.md` and `project-structure.md`

I would like to move the stuff in `src/trpc`, but I want to first figure out how our tRPC is even configured and learn how to do prefetching. So I will postpone this till another PR.

### Testing

It builds. I haven't exhaustively tested the entire website, but functionally nothing should have broke. We prob should test the entire website though.

Need to test:
- `migratePageViews.ts`

### AI Disclosure

Used for research: VS Code config structure, how tRPC works (mostly used their documentation tho)

### Checklist

- [x] Create this PR
- [x] Evaluate this PR against the [PR checklist](https://github.com/UTDNebula/engineeringSandbox/wiki/How-to-Contribute)
- [x] Perform final self-review

PR #732, opened 2026-08-29
Fix email contact schema regression

Isoscelestial:
Accidentally made it in #716 so you couldn't enter emails into the email contact field, this fixes that.


AbhiramTadepalli (approved):
LGTM!

PR #733, opened 2026-08-29
Make /events/calendar show all events

### LINKED ISSUE #714: View all events in event calendar

Viewing event calendar at `/events/calendar` should show all events, and should be accessible when signed out

### What happened?

Have to sign in to view the page, and it only shows events I've registered for (which is redundant with `/community/calendar`)

Preferred solution: Move the query outside of `/src/components/community/EventCalendar.tsx` so you can pass any events data you want into that component.

### Version

dev.clubs.utdnebula.com

_No response_

### Relevant log output

```shell

```

### Code of Conduct

Isoscelestial:
Resolves #714 

Makes `/events/calendar` page show events from all clubs instead of just registered events, and no longer requires logging in


AbhiramTadepalli (dismissed):
LGTM

PR #739, opened 2026-08-29
Auto-detect supported platforms for Other contact links

### LINKED ISSUE #736: Automatically convert an "Other" contact link to one of our pre-supported types
# Overview

We have these options for adding a contact link in a club listing:

Image

# Proposal
Detect if a url entered as "Other" matches any of the **un-filled** link options. If so, store it as that option instead of "Other".

## Supplementary Information
Skip email

lKolabrodl:
## Summary

- Detect supported contact platforms from URLs submitted as `Other`.
- Convert only into an unfilled platform, including `Website` as the HTTP(S) fallback, while leaving email detection disabled.
- Preserve existing contacts by replacing an existing `Other` row through the delete/create path when its database key changes.
- Keep contact ordering stable after a conversion.
- Add focused domain, lookalike-domain, occupied-platform, Website-fallback, and email-skip regression tests.

Closes #736

## Problem and user impact

Club managers can paste a Discord, Instagram, LinkedIn, or other recognized link into `Other`. That stores the wrong platform and gives the public listing a generic label/icon even when the matching platform slot is unused.

## Implementation

The new detector parses the URL hostname and matches exact domains or their subdomains. This avoids false positives such as `notinstagram.com` and query-string matches. A generic HTTP(S) URL falls back to `Website`; non-web schemes such as `mailto:` remain `Other`.

Before submission, `Other` entries are converted only if the inferred platform is not already present. Existing rows use `platform + clubId` as their key, so a converted stored row is deleted under `other` and recreated under the inferred platform. The submitted order uses the inferred key.

The Jest transform now sets an explicit `rootDir`, which TypeScript 6 requires in order to execute the new regression test (and the existing suite).

## Validation

- `npx jest tests/contactPlatform.test.ts --runInBand` — 18 tests passed
- `npm run lint:check`
- `npm run format:check`
- `SKIP_ENV_VALIDATION=1 npm run type:check`
- `SKIP_ENV_VALIDATION=1 npm run build` — application compilation and TypeScript passed; page-data collection then required the unavailable local `DATABASE_URL`

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -7,6 +7,9 @@ const config: JestConfigWithTsJest = {
     '^.+\\.[tj]sx?$': [
       'ts-jest',
       {
+        tsconfig: {
+          rootDir: '.',
+        },
         useESM: true,
       },
     ],
--- a/src/systems/manage/forms/Contacts/Contacts.tsx
+++ b/src/systems/manage/forms/Contacts/Contacts.tsx
@@ -30,6 +30,7 @@ import type { SelectClub, SelectContact } from '@/server/db/models';
 import { contactNames, startContacts } from '@/server/db/schema/contacts';
 import { useTRPC } from '@/trpc/react';
 import ContactListItem from './ContactListItem';
+import { convertOtherContactPlatforms } from './contactPlatform';
 import { editClubContactSchema } from './schema';
 
 type FormData = z.infer<typeof editClubContactSchema>;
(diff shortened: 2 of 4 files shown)


Isoscelestial:
Thank you for your contribution! However, we've closed this PR for at least one of these reasons:

- The issue this PR resolves is reserved for new recruits. You should instead find issues to work on using [these search filters](https://github.com/UTDNebula/utd-clubs/issues?q=is%3Aissue%20state%3Aopen%20-label%3A%22Recommended%20for%20Recruits%22).
- This PR uses AI in a way that breaks the [Nebula Labs AI Policy](https://nebula-labs.atlassian.net/wiki/x/AgCwQw). Please make a new PR with less AI usage.
- We plan to approach this issue in another way.

Please let us know if you have any questions!

PR #768, opened 2026-09-06
perf: stabilize useIsMounted subscription

### LINKED ISSUE #756: Slightly improve performance for useIsMounted hook
### Problem

In general, you shouldn't write an anonymous function for subscribe in usages of `useSyncExternalStore()` because it causes React to resubscribe each render, which slightly impacts performance.

In `src/lib/utils/useIsMounted.ts`:
```ts
export function useIsMounted(): boolean {
  const isMounted = useSyncExternalStore(
    () => () => {},
    () => true,
    () => false,
  );

  return isMounted;
}
```

### Solution

Instead of making it a completely anonymous function (i.e. a function without a name, which means they are created and immediately destructed), store the `() => () => {}` function in a variable first. This makes it a named function, which means it isn't different between renders.

### Additional Context

More info: https://discord.com/channels/860405177865338890/949773852211302451/1541275667704975420

Potentially relevant documentation page: https://react.dev/reference/react/useSyncExternalStore#my-subscribe-function-gets-called-after-every-re-render

SinhSinhAn:
## Summary

- move the no-op subscription function to module scope
- keep its reference stable across renders so useSyncExternalStore does not resubscribe unnecessarily

## Validation

- Prettier check on src/lib/utils/useIsMounted.ts
- ESLint on src/lib/utils/useIsMounted.ts
- SKIP_ENV_VALIDATION=1 npm run type:check

Fixes #756

### DIFF
--- a/src/lib/utils/useIsMounted.ts
+++ b/src/lib/utils/useIsMounted.ts
@@ -1,5 +1,7 @@
 import { useSyncExternalStore } from 'react';
 
+const subscribe = () => () => {};
+
 /**
  * Hook to return whether the page is finished mounting on the client.
  * Use for conditional components and props instead of
@@ -11,7 +13,7 @@ import { useSyncExternalStore } from 'react';
  */
 export function useIsMounted(): boolean {
   const isMounted = useSyncExternalStore(
-    () => () => {},
+    subscribe,
     () => true,
     () => false,
   );


Isoscelestial (approved):
lol thanks Sinh Sinh

It'd probably be good if you avoided the Recommended for Recruit issues I just created, gotta make sure new recruits have stuff to do and you'd probably do better with more difficult issues

also use the PR template next time

SinhSinhAn:
I stubbed my toes reading this, fsfs

PR #795, opened 2026-09-19
Email login on develop, env-based feature flags

### LINKED ISSUE #705: Implement Vercel feature flags
## Overview
Because we are now doing themed releases (see [Milestones](https://github.com/UTDNebula/utd-clubs/milestones)), we may have features that are semi-ready yet shouldn't be released publicly yet. Thus, we need [Vercel's feature flags](https://vercel.com/docs/flags/vercel-flags).

First usage of this will be for #622 

## Requirements
- Allow enabling/disabling features via Vercel
- Allow flags to be toggled for local development
- Add documentation on how to add feature flags

Isoscelestial:
### Description

Resolves #705 but mostly a bunch of other stuff

- Add email login
- Add "Manage Account" panel in account settings to change email and password (and also rework account deletion)
- Add feature flags system so that the above features are only available on development branches
  - Supports environment variables and defaultValue. Initially also supported Vercel via an adapter, but I realized that Vercel Flags cost money so I'm waiting to make that decision later.
- Change design of profile drop down when signed out, plus make it load a LOT faster
- Make provider buttons on login modal grayed out if environment variables are missing
- Adds a *few* environment variables (added a section to `.env.example`)

As an aside, I had a LOT of fun learning and putting to use some advanced React concepts like React Server Functions and Suspense.

Adds "feature flags" in the form of environment variables (the system does support Vercel Flags, but I removed it while developing because [Vercel Flags are paid](https://vercel.com/docs/flags/vercel-flags/limits-and-pricing) and this is something to discuss @AbhiramTadepalli):
- `email-auth` (`FLAG_EMAIL_AUTH`) - Controls whether visitors can sign in or sign up to accounts that use an email and password
  - `false` on prod, `true` on dev
- `password-requirements` (`FLAG_PASSWORD_REQUIREMENTS`) - Controls whether account passwords should have at least 3 of the following: lowercase, uppercase, number, symbol. Passwords must always have at least 8 characters regardless of this flag
  - `true` on prod, `false` on dev
- `login-banner-text` (`FLAG_LOGIN_BANNER_TEXT`) - Defines text to show in a banner on the login modal. Hides the banner if an empty string
  - Blank on prod, `"For now, email login is only available for local development"` on dev. This is just in case we need to quickly push a message to prod. Also supports markdown

### Testing

Normally the flags are disabled, but I added the necessary variables to Vercel for this particular preview: https://clubs-git-feat-email-login-on-development-utdnebula.vercel.app/

Locally, I tested it with:

- Test setting flags through environment variables
  - Should affect login modal, `/auth`, and `/settings`
- Test creating email account
- Test logging out of account
- Testing signing back into said account
- Test changing email and password and deleting account
- Tested that Google Calendar sync will at least show the calendars (i.e. that the access token works)
  - Too lazy rn get port forwarding working to test Google Calendar sync

### AI Disclosure

Helped me write some types for TypeScript

### Checklist

- [x] Create this PR
- [x] Perform final self-review


src/lib/utils/flags.ts
```diff
@@ -0,0 +1,255 @@
...
+}
+
+const logFlagOrigins = truthy(process.env.LOG_FLAG_ORIGINS ?? false);
+
+type DeciderFactoryOptions = {
+  /**
+   * Rather than obtaining the flag's value from environment variables first, prioritize obtaining the flag's value from the adapter
+   * @default false
+   */
+  prioritizeAdapter?: boolean;
```
AbhiramTadepalli: as of now, this isn't being used at all right?

src/server/auth.ts
```diff
@@ -18,8 +18,12 @@ const AUTH_TRUSTED_ORIGINS = Array.isArray(env.AUTH_TRUSTED_ORIGINS)
  */
 export const auth = betterAuth({
   database: drizzleAdapter(db, {
-    provider: 'pg', // or "pg" or "mysql"
+    provider: 'pg',
   }),
+  emailAndPassword: {
+    enabled: true,
+    requireEmailVerification: false, // TODO: add email verification for prod
```
AbhiramTadepalli: what happens when this is set to true?

justinschwerdtfeger:
just did a bit of testing, it doesnt seem like the flag to have password requirements can be set to false, its still requiring 8 characters

perhaps this flag could be called FLAG_STRICT_PASSWORD_REQUIREMENTS or something?

justinschwerdtfeger:
also I need to test to make sure gcal sync works

Ilutara (dismissed):

justinschwerdtfeger:
Just tested gcal sync, and it seems to work!
I still have a few more things i want to test with it

